PR #17, opened 2020-06-02
Improved subgraph template

ajsantander:
## Renaming
 
*packages/connect-thegraph-app-template* to
*packages/connect-thegraph-template*

## Completely new subgraph

Within the package, it also completely replaces the *packages/connect-thegraph-template/subgraph* sub-package.

This is a new template that not only serves as a subgraph template for Aragon apps, but for any Aragon related subgraph. It is more sophisticated in the way it manages data sources and supports more sources out of the box.

## Manifest file generation

It does not really change, but merely extends the way in which the subgraph.yaml manifest file is generated. The previous template only used 4 DAOFactory data sources, while this also uses DAO templates and the old DAO kits. It also allows the developer to set individual data sources for tokens and token factories.

## Redundancy management

Having so many data sources, this template now manages redundancy. For example, an organization may be picked up from a DAOFactory data sources and from a DAO template data source. We can't just create a template in both cases, since that would result in duplicate data.

Unfortunately, TheGraph doesn't have a way to check if a dynamic data source has already been created. The solution implemented here is to use a single *AragonInfo* entity that stores several arrays which are used to track the existence of particular data sources. Before creating a new template for a dynamic data source that has been found, we first check if it hasn't already been created using this entity.

```
type AragonInfo @entity {
  id: ID!
  orgs: [Bytes!]!
  apps: [Bytes!]!
  tokens: [Bytes!]!
}
```

Implementations of the template may also extend this entity to store additional global properties of a subgraph.

This entity is managed in src/aragon/aragon.ts.

## Modified hooks

This also introduces minor changes to the way hooks work. They are now called *aragon-hooks* and now also has hooks for org and token template creation.

For more info, pls see the subgraph/README.md file.

## After this PR

* Modify the voting and token-manager subgraphs to build on top of this template instead of the old one.


packages/connect-thegraph-template/subgraph/src/aragon/aragon.ts
```diff
@@ -0,0 +1,91 @@
...
+    aragon.tokens = entities
+  } else {
+    throw new Error('Invalid entity type ' + type)
+  }
+
+  aragon.save()
+}
+
+function _getAragonInfo(): AragonInfoEntity {
+  let aragonId = 'THERE_CAN_ONLY_BE_ONE'
```
Evalir, line 77: What exactly is this doing?

packages/connect-thegraph-template/subgraph/src/aragon/mappings/MiniMeToken.ts
```diff
@@ -0,0 +1,9 @@
+import { ClaimedTokens as ClaimedTokensEvent, NewCloneToken, Approval } from '../../../generated/templates/MiniMeToken/MiniMeToken'
+import { Transfer as TransferEvent } from '../../../generated/templates/MiniMeToken/MiniMeToken'
+import { NewCloneToken as NewCloneTokenEvent } from '../../../generated/templates/MiniMeToken/MiniMeToken'
+import { Approval as ApprovalEvent } from '../../../generated/templates/MiniMeToken/MiniMeToken'
+
+export function handleClaimedTokens(event: ClaimedTokensEvent): void {}
+export function handleTransfer(event: TransferEvent): void {}
+export function handleNewCloneToken(event: NewCloneTokenEvent): void {}
+export function handleApproval(event: ApprovalEvent): void {}
```
Evalir, line 9: Don't we have any way to handle these tokens events (and don't we have to handle them)?

sohkai:
> Unfortunately, TheGraph doesn't have a way to check if a dynamic data source has already been created

Can't we use the [`load()` trick](https://www.youtube.com/watch?v=JNqN3fek6FY&feature=youtu.be&t=340) here? Or are we unable to assume unique ids from multiple data sources?

packages/connect-thegraph-template/subgraph/src/aragon/aragon.ts
ajsantander: That's the ID of the AragonInfo entity. Since it's a sort of singleton, I wanted to give it an ID that makes it clear that there can only be one. Also, I watched highlander: https://www.youtube.com/watch?v=sqcLjcSloXs

packages/connect-thegraph-template/subgraph/src/aragon/mappings/MiniMeToken.ts
ajsantander: What do you mean? The way to handle those events is to enter some code on those functions. And no, you're not obligated to handle them.

ajsantander:
> Can't we use the load() trick here? Or are we unable to assume unique ids from multiple data sources?
@sohkai Unfortunately not. This is something I discussed with TheGraph's tech support. Load can be used on entities, but not on data sources.

packages/connect-thegraph-template/subgraph/abis/DAOKit.json
```diff
@@ -0,0 +1,293 @@
+[
```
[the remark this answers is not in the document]
ajsantander: ABIs: I'm just more comfortable with using whatever ABI etherscan shows when I go to the address of a given data source. Facu recommended to fetch or generate the ABIs ourselves, but it would be really complicated here because we would have to add a lot of dependencies.

Build: Build files should not be uploaded to git. If they are, it's an error.

packages/connect-thegraph-template/subgraph/src/aragon/mappings/MiniMeToken.ts
Evalir: Gotcha; I wanted to know if it was best to handle these, rather than leaving them unhandled by default.

ajsantander: Yeah I understand your concern. It wouldnt make sense to handle them in the template and in the Voting example that the template presents. We could comment out the code, but we would also need to do that in the yaml data sources, which would be awkward if the user wanted to enable them. That's why I decided to compromise with just leaving them empty. You can always optimize a subgraph by commenting them out in its implementation.

packages/connect-thegraph-template/subgraph/.gitignore
```diff
@@ -59,3 +59,6 @@ typings/
 
 # next.js build output
 .next
+
+# TheGraph artifacts
+generated
```
0xGabi, line 64: Maybe we should include the `build` folder to the ignored files?

packages/connect-thegraph-template/subgraph/subgraph.template.yaml
```diff
@@ -1,17 +1,20 @@
...
-{{/KernelDataSources}}
+{{> Organizations.yaml}}
+{{> OrganizationFactories.yaml}}
+{{> OrganizationTemplates.yaml}}
+{{> LegacyOrganizationTemplates.yaml}}
+{{> TokenFactories.yaml}}
+{{> Tokens.yaml}}
 templates:
   {{> Kernel.template.yaml}}
+  {{> MiniMeToken.template.yaml}}
```
0xGabi, line 17: Maybe include `Kernel.template.yaml` with ! to show that can be used as a template as well?

packages/connect-thegraph-template/subgraph/.gitignore
ajsantander: Yes! Thank you.

packages/connect-thegraph-template/subgraph/subgraph.template.yaml
ajsantander: Not sure what you mean. If what you mean is this:

```
templates:
  {{!> Kernel.template.yaml}}
  {{> MiniMeToken.template.yaml}}
```

That would disable the dynamic Kernal data source, which would break things.

PR #19, opened 2020-06-03
Updates voting and tokens subgraphs to use new template

ajsantander:
This PR simply updates our voting and token manager subgraphs to build on top of the new template proposed in https://github.com/aragon/connect/pull/17

An interesting thing to notice is that the token-manager subgraph is simplified, in the sense that it no longer needs to forcefully create the minime token template.


packages/connect-thegraph-token-manager/subgraph/src/TokenManager.ts
```diff
@@ -1,42 +1,30 @@
 import { Address } from '@graphprotocol/graph-ts'
 import { TokenManager as TokenManagerContract } from '../generated/templates/TokenManager/TokenManager'
 import { TokenManager as TokenManagerEntity } from '../generated/schema'
-import { createMiniMeTokenTemplateAndEntity } from './MiniMeToken'
 // import { NewVesting as NewVestingEvent } from '../generated/templates/TokenManager/TokenManager'
```
0xGabi, line 4: Do you leave this commented to improve the sync time? Should we include a comment about that?

ajsantander: Yep, that's the point. I'll add a comment!

PR #20, opened 2020-06-05
Implements pagination in thegraph connectors

ajsantander:
_Note_: Files changed show > 40, but that's only because I renamed connector-thegraph/subgraph-org, to connector-thegraph/subgraph so that it is consistent with other thegraph connectors. The actual changed files are very few.

#### Problem:

TheGraph returns 100 results max for any query by default. It can be instructed to return up to a 1000 using pagination "first" and "skip" parameters. None of the queries in our subgraphs are using these variables at all, thus returning the first 100 entities, if existent.

#### Solution:

After some analysis, I decided to only apply pagination in:
* Voting.votes()
* vote.casts()
* tokens.holders()

Here's why...

Yes, the subgraphs that the connectors use are capable of returning > 1000 results. For example, asking the tokens subgraph for the existing minime tokens can return thousands of results.

However, we actually don't use the subgraphs in this way in the API. Our queries tend to filter for a specific organization or a specific app. Considering this, which queries could return > 1000 results?

* permissions - extremely rare
* apps - extremely rare
* roles - extremely rare
* votes - rare but possible
* casts - rare but possible
* token holders - rare but possible

Modifying *all* our queries to be able to accept a { first, skip } object is overkill. Our low level GraphQLWrapper allows you to specify your query with these options in case you ever need them, but I think that this will be very rare.


examples/organization-viewer-cli/src/index.ts
```diff
@@ -134,7 +134,7 @@ async function inspectVotingHighLevel(appAddress: string): Promise<void> {
   console.log(voting.toString())
 
   console.log('\nVotes:')
-  const votes = await voting.votes()
+  const votes = await voting.votes(2, 1)
```
Evalir, line 137: Hmm, for some reason I naturally expect the `first` and `skip` params to be parts of an `options` object, e.g: `await voting.votes({ 2, 1})`. Not uncomfortable with the current implementation, but just a thought!

Evalir:
> Modifying all our queries to be able to accept a { first, skip } object is overkill. Our low level GraphQLWrapper allows you to specify your query with these options in case you ever need them, but I think that this will be very rare.

I think that we should have some documentation on doing this if users ever have the need to.

PR #21, opened 2020-06-05
Implements low-level subscriptions in thegraph connectors

ajsantander:
This PR implements the low level `subscribeToQuery(...)` function in **connect-thegraph/src/core/GraphQLWrapper.ts**.

See ` inspectVotingLowLevelSubscription(...)` in **examples/organization-viewer-cli/src/index.ts** for usage.
If a vote is created in the org while the demo is running, you will see it pop up in the demo's stdout.

_Note_: Next steps regarding subscriptions is to actually use this low level function in the connectors. This will probably imply slight modifications to how we build gql queries, and an event system that connectors must comply to (`on(...)`, `off(...)`), but not much more than that to have working subscriptions all over.

### DIFF
--- a/packages/connect-thegraph/src/core/GraphQLWrapper.ts
+++ b/packages/connect-thegraph/src/core/GraphQLWrapper.ts
@@ -1,20 +1,61 @@
 import fetch from 'isomorphic-unfetch'
-import { Client } from '@urql/core'
+import ws from 'ws'
+import { Client, defaultExchanges, subscriptionExchange, createRequest } from '@urql/core'
+import { SubscriptionClient } from 'subscriptions-transport-ws'
 import { DocumentNode } from 'graphql'
 import { ParseFunction, QueryResult } from '../types'
+import { pipe, subscribe } from 'wonka'
+
+type SubscriptionCallback = (result: any) => void
+type Subscription = {
+  unsubscribe: () => void
+}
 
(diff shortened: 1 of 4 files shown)


packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -1,30 +1,78 @@
...
+    )
+
     this.#client = new Client({
       maskTypename: true,
       url: subgraphUrl,
       fetch,
+      exchanges: [
+        ...defaultExchanges,
+        subscriptionExchange({
+          forwardSubscription: (operation: any) => subscriptionClient.request(operation)
```
0xGabi, line 35: Wonder if `graphql` or `urql` might have support for the operation type instead of using any here.

packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -1,30 +1,78 @@
...
+  subscribeToQuery(
+    query: DocumentNode,
+    args: any = {},
+    callback: SubscriptionCallback
+  ): Subscription {
+    const request = createRequest(query, args)
+
+    return pipe(
+      this.#client.executeSubscription(request),
+      subscribe((result: any) => {
```
0xGabi, line 52: Same comment here, wonder if `result` might have a type we could use.

PR #23, opened 2020-06-09
Implements high level subscriptions

ajsantander:


examples/organization-viewer-cli/package.json
```diff
@@ -7,7 +7,13 @@
     "build": "yarn clean && yarn compile",
     "dev": "yarn build --watch",
     "clean": "rm -rf ./dist ./tsconfig.tsbuildinfo",
-    "start": "node dist/src/index.js",
```
ajsantander, line 10: I split the cli demo into multiple demos because it was becoming unmanageable. I like the new separation a lot because going into each specific demo makes it very clear. It should also allow us to further simplify them because their setup no longer needs to be so generic.

packages/connect-core/src/connections/ConnectorInterface.ts
```diff
@@ -11,7 +11,9 @@ import Role from '../entities/Role'
 export interface ConnectorInterface {
   chainId?: number
   permissionsForOrg(orgAddress: string): Promise<Permission[]>
```
ajsantander, line 13: Following are the new functions that core level connectors need to implement. Notice how:
1. It uses separate functions for subscriptions
2. Some single-call interface functions do not have a subscription analogy. Thought it doesn't make sense in some cases.

packages/connect-core/src/entities/Organization.ts
```diff
@@ -48,6 +48,10 @@ export default class Organization {
     return this.#connector.appsForOrg!(this.#address)
   }
```
ajsantander, line 50: This is how an entity implements a subscription function. It's basically a mapping to the connector's corresponding function, much like it already was with single-call functions.

packages/connect-thegraph/src/connector.ts
```diff
@@ -39,58 +36,60 @@ export default class ConnectorTheGraph extends GraphQLWrapper
...
   async rolesForAddress(appAddress: string): Promise<Role[]> {
-    const result = await this.performQuery(queries.ROLE_BY_APP_ADDRESS, {
-      appAddress: appAddress.toLowerCase(),
-    })
-
-    const datas = this.parseQueryResult(parseRoles, result)
-
-    return datas.map((data: RoleData) => {
-      return new Role(data, this)
-    })
```
ajsantander, line 50: I created a new low level function `performQueryWithParser(...)` which simplifies connector functions. Parsers now return the final objects, so all the parsing is delegated to the parsers, as it should.

packages/connect-thegraph/src/connector.ts
```diff
@@ -39,58 +36,60 @@ export default class ConnectorTheGraph extends GraphQLWrapper
...
+    return this.performQueryWithParser(
+      queries.ORGANIZATION_PERMISSIONS('query'),
+      { orgAddress: orgAddress.toLowerCase() },
+      parsePermissions
+    )
+  }
 
-    return datas.map((data: PermissionData) => {
-      return new Permission(data, this)
-    })
```
ajsantander, line 62: And `subscribeToQueryWithParser(...)` is `performQueryWithParser(...)`'s analogy function with subscriptions. With this, the connector looks very straight forward and is easy to digest.

packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -7,11 +7,6 @@ import { DocumentNode } from 'graphql'
 import { ParseFunction, QueryResult } from '../types'
 import { pipe, subscribe } from 'wonka'
```
ajsantander, line 9: I removed these types so that they can be handled in a more generic way, without requiring users to find obscure types. They're very intuitive imo. We can add them back if you don't like it.

packages/connect-thegraph/src/parsers/apps.ts
```diff
@@ -1,10 +1,13 @@
-import { AppData } from '@aragon/connect-core'
```
ajsantander, line 1: Following are the changes to the parsers. Remember, they're returning entities now, not just data objects.

packages/connect-thegraph/src/queries/index.ts
```diff
@@ -1,8 +1,8 @@
 import gql from 'graphql-tag'
 import * as fragments from './fragments'
```
ajsantander, line 3: Queries are no longer constants, but instead they are functions that return a gql tag. This is because they can now be used with `query` or `subscription`. The type is not being enforced, so connectors should be careful about using the correct strings. I thought that enforcing it would be overkill, but it can be added if you like.

examples/organization-viewer-cli/package.json
0xGabi: Nice, good idea!

packages/connect-thegraph/src/connector.ts
0xGabi: Great abstraction

packages/connect-thegraph/src/parsers/permissions.ts
```diff
@@ -30,4 +33,10 @@ export function parsePermissions(result: QueryResult): PermissionData[] {
       }
     }
   )
+
+  const allowedPermissions = datas.filter((data: PermissionData) => data.allowed)
```
0xGabi, line 37: Much better to handle it here 💪

packages/connect-thegraph/src/queries/index.ts
0xGabi: Nice, I agree

PR #27, opened 2020-06-11
Radspec descriptions

0xGabi:
Implementation of Radspec descriptions. This PR is using a new version of Radspec that uses Ethers betav5 implemented here: https://github.com/aragon/radspec/pull/100

The changes include Radspec descriptions without considering more complex use cases like Intent Basket that we don't support yet on Connect. I include the changes to the main Radspec logic in a specific commit to make it easier to review: https://github.com/aragon/connect/pull/27/commits/5e166a4d6089a92f6eedb73f89edf086a0647ee5

The changes can be tested with the transaction-path cli example.

I'm leaving a few comments that I'll appreciate your suggestions.


packages/connect-core/package.json
```diff
@@ -19,7 +19,8 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
-    "ethers": "^5.0.0-beta.188"
+    "ethers": "^5.0.0-beta.188",
+    "radspec-gabi": "^2.0.1"
```
0xGabi, line 23: After a final round of review on the PR on Radspec we should publish the official version and update here.

packages/connect-core/src/entities/Permission.ts
```diff
@@ -27,7 +27,12 @@ export default class Permission extends CoreEntity implements PermissionData {
   constructor(data: PermissionData, connector: ConnectorInterface) {
     super(connector)
 
-    Object.assign(this, data)
```
0xGabi, line 30: Removing all `Object.assign` because they broke the type safety.

packages/connect-core/src/transactions/TransactionRequest.ts
```diff
@@ -22,6 +29,8 @@ export default class TransactionRequest {
     from: string
     data: string
   }
+  readonly description?: string
```
0xGabi, line 32: I decided to expand the `TransactionRequest` object to include both the `description` and `annotatedDescription` values. What do you think? I wonder if this might introduce problems when interacting with other libraries and we should handle it as a separate `TransactionDescription` object.

packages/connect-core/src/utils/path/describePath.ts
```diff
@@ -0,0 +1,46 @@
...
+  tryEvaluatingRadspec,
+  postprocessRadspecDescription,
+} from '../radspec/index'
+import { TransactionRequestData } from '../../transactions/TransactionRequest'
+
+/**
+ * Use radspec to create a human-readable description for each transaction in the given `path`
+ *
+ */
+export async function describeTransactionPath(
```
0xGabi, line 15: This function is only considering the straight forward Radspec implementation. Here we will include a couple more handlers when we improve our transaction path support.

packages/connect-core/src/utils/radspec/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    },
+    undefined
+  )
+
+  const { abi, method } = foundMethod || {}
+
+  let evaluatedNotice
+  if (method && method.notice) {
+    try {
+      evaluatedNotice = await radspec.evaluate(
```
0xGabi, line 53: Here is where we call the Radspec library.

packages/connect-core/src/utils/radspec/postprocess.ts
```diff
@@ -0,0 +1,126 @@
+import { addressesEqual, ANY_ENTITY } from '../address'
```
0xGabi, line 1: This is a post-process logic, pretty much the same we are using in the wrapper.

packages/connect-core/src/utils/radspec/postprocess.ts
```diff
@@ -0,0 +1,126 @@
...
+import { getKernelNamespace } from '../kernel'
+import Application from '../../entities/Application'
+import Role from '../../entities/Role'
+import { Annotation } from '../../transactions/TransactionRequest'
+
+export interface PostProcesDescription {
```
Evalir, line 7:
```suggestion
export interface PostProcessDescription {
```

packages/connect-core/src/utils/radspec/postprocess.ts
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Look for known addresses and roles in a radspec description and substitute them with a human string
+ *
+ * @param  {string} description
+ * @return {Promise<Object>} Description and annotated description
+ */
+export async function postprocessRadspecDescription(
+  description: string,
+  apps: Application[]
+): Promise<PostProcesDescription> {
```
Evalir, line 28:
```suggestion
): Promise<PostProcessDescription> {
```

packages/connect-core/src/transactions/TransactionRequest.ts
0xGabi: No, I just wonder if having these extra parameters could provoke problems when sending the transaction through Ethers or Web3.

packages/connect-core/src/types.ts
```diff
@@ -1,31 +1,8 @@
...
-  type: string
-  name: string
-  constant: boolean
-  inputs: Array<ParamType>
-  outputs: Array<ParamType>
-  payable: boolean
-  stateMutability: string
-  gas?: string
-}
+export type FunctionFragment = ethers.utils.FunctionFragment
```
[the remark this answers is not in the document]
0xGabi: You are right, this was because I add the above type later and forgot to update here.

packages/connect-core/src/types.ts
```diff
@@ -40,7 +17,7 @@ export interface AppIntent {
    * The function's ABI element is included for convenience of the client
    * null if ABI is not found for this signature
    */
-  abi: FunctionFragment | null
+  abi: ethers.utils.FunctionFragment | null
```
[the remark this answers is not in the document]
0xGabi: You are right, I define the type later and forgot to update here.

PR #33, opened 2020-06-12
Adds tests for core thegraph connector

ajsantander:
Not sure about the added `packages/test-helpers`

### DIFF
--- a/packages/connect-thegraph/src/__test__/permissions.test.ts
+++ b/packages/connect-thegraph/src/__test__/permissions.test.ts
@@ -1,0 +1,55 @@
+import ConnectorTheGraph from '../connector'
+import {
+  Permission
+} from '@aragon/connect-core'
+import {
+  isAddress,
+  isBytes32
+} from '../../../test-helpers'
+
+const DAO_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/aragon/aragon-mainnet'
+const ORG_ADDRESS = '0x0c188b183ff758500d1d18b432313d10e9f6b8a4'
+
+describe('when connecting to the mainnet subgraph', () => {
+  let connector: ConnectorTheGraph
+
+  beforeAll(() => {
+    connector = new ConnectorTheGraph({
+      daoSubgraphUrl: DAO_SUBGRAPH_URL
+    })
+  })
+
+  describe('when querying for the permissions of an org', () => {
+    let permissions: Permission[]
+
+    beforeAll(async () => {
+      permissions = await connector.permissionsForOrg(ORG_ADDRESS)
+    })
+
+    test('permissions are found', () => {
+      expect(permissions.length).toBeGreaterThan(0)
+    })
+
+    describe('when looking at a permission', () => {
+      let permission: Permission
+
+      beforeAll(() => {
+        permission = permissions[Math.floor(permissions.length / 2)]
+        // console.log(permission)
+      })
+
+      test('has a valid app address', () => {
+        expect(isAddress(permission.appAddress)).toBeTruthy()
+      })
+
+      test('has a valid grantee address', () => {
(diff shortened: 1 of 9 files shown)


packages/test-helpers/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './isAddress'
```
ajsantander, line 1: Here :point_down: 

Not sure if these should be a package or not. For now, it's just code that the other packages can import in their tests relatively.

packages/test-helpers/isAddress.ts
```diff
@@ -0,0 +1,3 @@
+export function isAddress(value: string): boolean {
+  return /^(0x)?[0-9a-f]{40}$/i.test(value)
```
bpierre, line 2: I don’t think we have addresses without `0x` prefix:

```suggestion
  return /^0x[0-9a-f]{40}$/i.test(value)
```

packages/test-helpers/isBytes32.ts
```diff
@@ -0,0 +1,3 @@
+export function isBytes32(value: string): boolean {
+  return /^(0x)?[0-9a-f]{64}$/i.test(value)
```
bpierre, line 2:
```suggestion
  return /^0x[0-9a-f]{64}$/i.test(value)
```

packages/test-helpers/index.ts
bpierre: I think it’s good enough for now, we can make it a package whenever it becomes needed. :+1:

packages/connect-thegraph/src/__test__/permissions.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    test('permissions are found', () => {
+      expect(permissions.length).toBeGreaterThan(0)
+    })
+
+    describe('when looking at a permission', () => {
+      let permission: Permission
+
+      beforeAll(() => {
+        permission = permissions[Math.floor(permissions.length / 2)]
+        // console.log(permission)
```
0xGabi, line 38: I think we forgot to remove this `console.log`.

packages/test-helpers/index.ts
0xGabi: We had a utility for `isAddress` that I decide to remove because ethers support it. I review and have both: [`isAddress`](https://docs.ethers.io/v5/api/utils/address/#utils-isAddress) and [`isBytes`](https://docs.ethers.io/v5/api/utils/bytes/#utils-isBytes).

Maybe we should use that?

ajsantander: I was trying to avoid using a library for now. Having a helpers package funnels everything to one location. You can go there and see if its got what you need instead of requiring you to know that Ethers can do it.

Let's keep it, and in the future _use_ Ethers from within the helpers when this becomes a package.

packages/connect-thegraph/src/__test__/permissions.test.ts
ajsantander: Linter coming soon! All these will be annihilated.

ajsantander:
@0xGabi yeah, saw that too. Let's merge and see if it can be fixed later.

> Seems that running the tests in parallel is the issue. If we use the ` --runInBand --detectOpenHandles` options as suggested by Jest on the `test:watch` script. The issue is gone. Regardless not sure yet why the issue is happening.

bpierre:
> Seems that running the tests in parallel is the issue. If we use the ` --runInBand --detectOpenHandles` options as suggested by Jest on the `test:watch` script. The issue is gone. Regardless not sure yet why the issue is happening.

Pretty sure it’s because we don’t disconnect: https://github.com/aragon/connect/blob/79cc64ee3c2b60ba97a5bcde8e4144f4b67df893/packages/connect-thegraph/src/core/GraphQLWrapper.ts#L15

PR #34, opened 2020-06-12
Adds some tests for voting

ajsantander:
Pls see comments on "Files changed" view

### DIFF
--- a/packages/connect-thegraph-voting/src/__test__/votes.test.ts
+++ b/packages/connect-thegraph-voting/src/__test__/votes.test.ts
@@ -1,0 +1,83 @@
+import { VotingConnectorTheGraph, Vote } from '../../src'
+
+const VOTING_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/aragon/aragon-voting-rinkeby'
+const VOTING_APP_ADDRESS = '0xc73e86aab9d232495399d62fc80a36ae52952b81'
+
+describe('when connecting to a voting app', () => {
+  let connector: VotingConnectorTheGraph
+
+  beforeAll(() => {
+    connector = new VotingConnectorTheGraph(
+      VOTING_SUBGRAPH_URL
+    )
+  })
+
+  describe('when querying for all the votes of a voting app', () => {
+    let votes: Vote[]
+
+    beforeAll(async () => {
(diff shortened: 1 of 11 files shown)


packages/connect-thegraph-voting/src/__test__/votes.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      test('should be executed', () => {
+        expect(vote.executed).toBe(true)
+      })
+
+      test('should have no nays', () => {
+        expect(vote.nay).toBe('0')
+      })
+
+      // TODO: Unskip when bug in subgraph is fixed
+      test.skip('should have yeas', () => {
```
ajsantander, line 45: There was a bug in the subgraph which caused this to fail. It's syncing now. Will update by removing this skip once it's done.

packages/connect-thegraph-voting/src/entities/Vote.ts
```diff
@@ -22,7 +21,6 @@ export default class Vote extends Entity implements VoteData {
   readonly id!: string
   readonly creator!: string
   readonly metadata!: string
-  readonly open!: boolean
```
ajsantander, line 25: I realized that the subgraph is not really capable of tracking this property since its not triggered by an event, but by a time out. There should be a way to determine it from connect, but I just disabled it for now.

packages/connect-thegraph-voting/subgraph/manifest/data/rinkeby-staging.json
```diff
@@ -8,28 +8,7 @@
         "startBlock": "6430210"
       }
     ],
-    "OrganizationFactories": [
```
ajsantander, line 11: These sources shouldn't really be in staging.

packages/connect-thegraph-voting/subgraph/src/Voting.ts
```diff
@@ -35,6 +35,12 @@ export function handleCastVote(event: CastVoteEvent): void {
   casts.push(castId)
   vote.casts = casts
 
+  if (event.params.supports == true) {
```
ajsantander, line 38: Yep! We weren't tracking yea/nay count at all.

PR #36, opened 2020-06-13
ENS support + a few other things

bpierre:
Add ENS support for `connect()`.

Also:

- Renamed `examples/react` into `examples/org-viewer-web`.
- Add a package for shared types: `@aragon/connect-types`.
- GraphQLWrapper: improve error messages.
- Voting demo: tweak output + dynamic network.
- Refresh the documentation, especially around `connect()`.
- `ConnectorTheGraph`: move from `daoSubgraphUrl` to `orgSubgraphUrl`.
- `ConnectorTheGraph`: pick the subgraph URL based on the `chainId`.
- Update the demos and tests for the new `connect()` format.


package.json
```diff
@@ -34,6 +25,7 @@
     "connect-thegraph": "yarn workspace @aragon/connect-thegraph",
     "connect-thegraph-voting": "yarn workspace @aragon/connect-thegraph-voting",
     "connect-thegraph-token-manager": "yarn workspace @aragon/connect-thegraph-token-manager",
+    "connect-types": "yarn workspace @aragon/connect-types",
```
Evalir, line 28: I think this is nice, but have we looked into possibly publishing individual types as per defined by [typescript](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html), in a `@types/aragon-connect` package?

examples/nodejs/src/voting.ts
```diff
@@ -1,21 +1,53 @@
...
+// const ALL_VOTING_SUBGRAPH_URL =
+//   'https://api.thegraph.com/subgraphs/name/aragon/aragon-voting-mainnet'
 
-async function main() {
-  console.log('\nVoting:')
+function voteTitle(vote: Vote) {
+  return vote.metadata
+    .split('\n')[0]
+    .replace(/\(Link[^\)]+\)/, '')
+    .replace(/\(SHA256[^\)]+\)/, '')
```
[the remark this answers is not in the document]
bpierre: Just some cleanup for the AGP votes :smile: https://mainnet.aragon.org/#/governance.aragonproject.eth/0x277bfcf7c2e162cb1ac3e9ae228a3132a75f83d4/

I added a comment :+1:

package.json
bpierre: We are already [publishing the types with our packages](https://github.com/aragon/connect/blob/master/packages/connect-core/package.json#L6), so I don’t think we need to publish on `@types/xyz`? Do you see a reason to do it?

The `@aragon/connect-types` package is meant to contain the types that are shared by our different packages, and its types to be exported by the different packages using them. We could also have them in `@aragon-core`, but having them in a separate package is something we discussed recently and wanted to try.

packages/connect-thegraph-voting/src/queries/index.ts
```diff
@@ -10,7 +10,6 @@ export const ALL_VOTES = (type: string) => gql`
       orgAddress
       creator
       metadata
-      open
```
bpierre, line 13: I am not sure if we should remove this? This is the error I get when leaving it:

> CombinedError: [GraphQL] Type `Vote` has no field `open`

package.json
Evalir: Yeah, indeed there shouldn't be a need to publish them now; it's probably something we could do later if we ever need to. Wonder if having the types in a separate package could cause issues with linting tools? Guess that's why we're trying haha!

bpierre: > Wonder if having the types in a separate package could cause issues with linting tools? Guess that's why we're trying haha!

Things seem to work up to this point, but let’s keep an eye on it yes! :smile:

packages/connect-thegraph-voting/src/queries/index.ts
0xGabi: Yes, we should. The latest version of the schema doesn't have that property.

packages/connect-core/src/entities/Organization.ts
```diff
@@ -25,35 +26,68 @@ import { ConnectorInterface } from '../connections/ConnectorInterface'
...
   ) {
-    this.#address = address
-    this.#provider =
-      provider ||
-      new ethers.providers.InfuraProvider(chainId || connector.chainId || 1)
+    this.location = location
+
+    this.#provider = provider
+      ? new ethers.providers.Web3Provider(provider, network)
+      : new ethers.providers.InfuraProvider(network)
```
0xGabi, line 46: What you think of instead using the `FallbackProvider`? I think we might see a better performance. For example, we can decide to use a weight of 0.5 for our own node and then distribute the other half equally between the API providers (EtherscanProvider, AlchemyProvider, InfuraProvider).

bpierre: Ah yes I was checking it recently, it sounds super interesting! The only thing is that it doesn’t seem to support a network configuration (i.e. a custom ENS address)?

0xGabi: Oh maybe is not possible but I thought that would be configurable using the: [FallbackProviderConfig](https://docs.ethers.io/v5/api/providers/other/#FallbackProviderConfig)

bpierre: Oooh I understand now, we pass a configuration for each of them. It should work then, as long as each provider we want also supports custom ENS addresses :+1:

0xGabi: At least is how I understood it, but we can give it a shoot for the next release for sure 🙌

PR #37, opened 2020-06-13
Adds tests for the tokens connector

ajsantander:


packages/connect-thegraph-token-manager/subgraph/src/aragon/aragon.ts
```diff
@@ -80,6 +80,7 @@ function _getAragonInfo(): AragonInfoEntity {
   if (!aragon) {
     aragon = new AragonInfoEntity(aragonId)
 
+    aragon.version = 'tokens-v0.1.0'
```
ajsantander, line 83: This might be interesting for you @0xGabi 

For now, I'm not doing anything with this. It's just in case I ever doubt the version that is currently synced. However, we could have connectors look for this info and complain if they find a version that they don't know how to handle, or is deprecated.

packages/connect-thegraph-token-manager/package.json
```diff
@@ -7,16 +7,23 @@
   "scripts": {
     "build": "yarn clean && yarn compile",
     "dev": "yarn build --watch",
+    "test": "is-ci \"test:coverage\" \"test:watch\"",
```
sohkai, line 10: Since we're using `is-cli` here, we may also want to mark it as a dev dependency

PR #43, opened 2020-06-16
Update transaction-path example and handle provider

0xGabi:


packages/connect-core/src/entities/Organization.ts
```diff
@@ -37,13 +37,21 @@ export default class Organization {
     location: string,
     connector: ConnectorInterface,
     provider: any,
-    network: Networkish
+    network: Network | string
   ) {
     this.location = location
 
+    const getProvider = (): ethers.providers.Provider => {
```
bpierre, line 44: Should we name this `getEthersProvider()`? To make it clear that it is not an [Ethereum provider](https://eips.ethereum.org/EIPS/eip-1193) (which ethers calls a “Web3 Provider”, to make it more confusing :laughing:).

packages/connect-core/src/entities/Organization.ts
```diff
@@ -37,13 +37,21 @@ export default class Organization {
     location: string,
     connector: ConnectorInterface,
     provider: any,
-    network: Networkish
+    network: Network | string
```
bpierre, line 40: Ah yes I forgot to make the `Networkish` => `Network` update here. :+1: 

About `string`, do you see a case where this would be needed?

0xGabi: I went with the default that Ethers asked for.

bpierre: I think Ethers wants a `Networkish`? https://github.com/ethers-io/ethers.js/blob/ethers-v5-beta/packages/providers/src.ts/web3-provider.ts#L81

But in the case of `Organization`, I think we can restrict it to `Network` to make it easier (since `connect()` is already doing the `Networkish` => `Network` conversion).

Issue #48, opened 2020-06-17
Add filters to org.app() and org.apps()


bpierre:
The `app()` and `apps()` methods of `Organization` could provide a way to filter the desired apps, by having a filtering object passed to them.

Proposed API:

```tsx
// Selecting by appName
org.apps({ appName: 'voting.aragonpm.eth' })

// Selecting multiple appName
org.apps({ appName: ['voting.aragonpm.eth', 'token-manager.aragonpm.eth'] })

// Selecting by address
org.apps({ address: '0xcafe…' })

// Selecting multiple addresses
org.apps({ address: ['0xcafe…', '0xbeef…'] })
```

We could have other filters but `address` and `appName` would probably enough for now.

Selecting the first app that matches could be done this way:

```tsx
const [voting] = await org.apps({ appName: 'voting.aragonpm.eth' })
```

Or by using `.app()`, which should act as a convenient alias for the previous example:

```tsx
const voting = await org.app({ appName: 'voting.aragonpm.eth' })
```

The only difference is that `.app()` would also accept an address to be passed directly:

```tsx
// these two lines are equivalent:
const voting = await org.app('0xcafe…')
const [voting] = await org.apps({ address: '0xcafe…' })
```

It will eventually make it possible to do this, for an org with a single voting app:

```tsx
const voting = new Voting(await org.app({ appName: 'voting.aragonpm.eth' }))
```

Note: this is not the case yet, but we sould try to reuse the names and patterns we use for filters as much as possible so it feels consistent.

PR #52, opened 2020-06-18
Connect for React API

bpierre:
_Builds on #111_

# Summary

This PR adds an initial version of the React library.

## &lt;Connect />

A decision was made to move the configuration on the `<Connect />` component, which makes it possible to declare it once in the app. The idea is to optimize for the use case of connecting to a single organization, and interacting with it. Users can still declare `<Connect />` in different places to connect to more than one organization, or use `@aragon/connect` directly.

## Hooks for subscriptions only

The only hooks provided are abstracting the subscriptions. Async calls are left as promises for now. This is something we might want to reconsider in the future depending on usage.

## @aragon/connect re-exports

The entire @aragon/connect is being exported by the React library. This is to make it easier when adding new types, but we might prefer to list the exports instead.

## App connectors

There is no hook allowing to subscribe to apps through their connectors yet. The plan is to do it as a second step, once they can be instantiated from the connection in the core library.

## Organization

The `Organization` object has been left as it is, including methods that are not needed with the react library like `.apps()`. This is to avoid having different types for the React library and for the others.

<hr />
# Using Aragon Connect with React

## Introduction

Aragon Connect provides a series of utilities that makes it easier to use in React environments.

It consists of the `<Connect />` component, on which a connection to an organization is described, and a series of hooks: `useApp()`, `useApps()`, `useOrganization()`, `usePermissions()`.

## Usage

```jsx
import {
  Connect,
  useApps,
  useOrganization,
  usePermissions,
} from '@aragon/connect-react'

function App() {
  const [org, orgStatus] = useOrganization()

  const [apps, appsStatus] = useApps()
  const [permissions, permissionsStatus] = usePermissions()

  const loading =
    orgStatus.loading || appsStatus.loading || permissionsStatus.loading
  const error = orgStatus.error || appsStatus.error || permissionsStatus.error

  if (loading) {
    return <p>Loading…</p>
  }

  if (error) {
    return <p>Error: {error.message}</p>
  }

  return (
    <>
      <h1>{org.name}</h1>

      <h2>Apps</h2>
      <ul>
        {apps.map((app, i) => (
          <li key={i}>{app.name}</li>
        ))}
      </ul>

      <h2>Permissions</h2>
      <ul>
        {permissions.map((permission, i) => (
          <li key={i}>{String(permission)}</li>
        ))}
      </ul>
    </>
  )
}

ReactDOM.render(
  <Connect location="myorg.aragonid.eth" connector="thegraph">
    <App />
  </Connect>,
  document.querySelector('main')
)
```

## API

### &lt;Connect />

This component is required in order to use the provided hooks.

| Props                  | Type                                      | Description                                                                                              |
| :--------------------- | :---------------------------------------- | :------------------------------------------------------------------------------------------------------- |
| `location`             | `String`                                  | The Ethereum address or ENS domain of an Aragon organization.                                            |
| `connector`            | `Connector \| [String, Object] \| String` | Accepts a `Connector` instance, and either a string or a tuple for embedded connectors and their config. |
| `options`              | `Object`                                  | The optional configuration object.                                                                       |
| `options.readProvider` | `EthereumProvider`                        | An [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) compatible object.                                |
| `options.chainId`      | `Number`                                  | The [Chain ID](https://chainid.network/) to connect to. Defaults to `1`.                                 |

### useOrganization()

| Props   | Type                                                                                  | Description                                                                                           |
| :------ | :------------------------------------------------------------------------------------ | :---------------------------------------------------------------------------------------------------- |
| returns | `[Organization \| null, { loading: boolean, error: null \| Error, retry: Function }]` | An array containing the [organization](../api-reference/organization.md) and a loading status object. |

### useApp(appFilters)

| Name                | Type                                                                         | Description                                                                                                                                                                                                                                                   |
| :------------------ | :--------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `appFilter`         | `String` or object (optional)                                                | When a string is passed, the app will get searched by address if it starts by `0x`, and by `appName` otherwise. See `appFilter.address` and `appFilter.appName` to set them explicitly. For the time being, only one type of filter can get passed at a time. |
| `appFilter.address` | `String`                                                                     | Same as `appFilter`, but makes the selection by `address` explicit.                                                                                                                                                                                           |
| `appFilter.appName` | `String`                                                                     | Same as `appFilter`, but makes the selection by `appName` explicit.                                                                                                                                                                                           |
| returns             | `[App \| null, { loading: boolean, error: null \| Error, retry: Function }]` | An array containing a single [app](../api-reference/app.md) from the organization and a loading status object.                                                                                                                                                |

### useApps(appFilters)

| Name                | Type                                                                   | Description                                                                                                                                                                                                                                                                                                                        |
| :------------------ | :--------------------------------------------------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `appFilter`         | `String` or `String[]` or object (optional)                            | When a string is passed, apps will get filtered by address if it starts by `0x`, and by `appName` otherwise. When an array is passed, the first entry determines the type of filter. See `appFilter.address` and `appFilter.appName` to set them explicitly. For the time being, only one type of filter can get passed at a time. |
| `appFilter.address` | `String` or `String[]`                                                 | Same as `appFilter`, but makes the selection by `address` explicit.                                                                                                                                                                                                                                                                |
| `appFilter.appName` | `String` or `String[]`                                                 | Same as `appFilter`, but makes the selection by `appName` explicit.                                                                                                                                                                                                                                                                |
| returns             | `[App[], { loading: boolean, error: null \| Error, retry: Function }]` | An array containing the organization [apps](../api-reference/app.md) and a loading status object.                                                                                                                                                                                                                                  |

### usePermissions()

| Name    | Type                                                                          | Description                                                                                                     |
| :------ | :---------------------------------------------------------------------------- | :-------------------------------------------------------------------------------------------------------------- |
| returns | `[Permission[], { loading: boolean, error: null \| Error, retry: Function }]` | An array containing the organization [permissions](../api-reference/permission.md) and a loading status object. |


sohkai:
> useApps()

What is the strategy for connecting to a particular app in a react context? Will that be the domain of the app connector to define?

bpierre:
> What is the strategy for connecting to a particular app in a react context? Will that be the domain of the app connector to define?

I am currently adding a `.connect()` method to the `App` object:

```jsx
import { connect } from '@aragon/connect'
import VotingConnector from '@aragon/connect-thegraph-voting'

const org = await connect('myorg.aragonid.eth', 'thegraph')

let votesSubscription

org.onApps(apps => {
  const voting = apps.find(app => app.appName.startsWith('voting.'))
  const votingConnector = await voting.connect(VotingConnector)

  votesSubscription?.unsubscribe()
  votesSubscription = votingConnector.onVotes(null, votes => {
    console.log(votes)
  })
})

```

Then I was thinking of adding the equivalent in React:

```jsx
import { useApps, useAppSubscription } from '@aragon/connect-react'
import VotingConnector from '@aragon/connect-thegraph-voting'

function App() {
  const [apps] = useApps()
  const voting = apps.find(app => app.appName.startsWith('voting.'))

  // We would take care of reusing the connection internally, for any given app + app connector.
  // Params are: app, app connector, method, method params.
  const votes = useAppSubscription(voting, VotingConnector, 'onVotes', {}) ?? []

  return (
    <ul>
      {votes.map(vote => (
        <li>{vote.id}</li>
      ))}
    </ul>
  )
}

export default () => (
  <Connect location="myorg.aragonid.eth" connector="thegraph">
    <App />
  </Connect>
)
```

Another option would be to reuse the [app filters](https://github.com/aragon/connect/issues/48), to let people declare them at the `Connect` level:

```jsx
import { useApps, useAppSubscription } from '@aragon/connect-react'
import VotingConnector from '@aragon/connect-thegraph-voting'

function App() {
  const [apps] = useApps()
  const voting = apps.find(app => app.appName.startsWith('voting.'))

  // No need to pass the connector here, it’s already set by Connect.
  const votes = useAppSubscription(voting, 'onVotes', {}) ?? []

  // It could also be a good idea to extend App with a connector property.
  // We might also want to consider having a similar system on the core library,
  // so that users could pass all their app connectors at the connect() level. 

  return (
    <ul>
      {votes.map(vote => (
        <li>{vote.id}</li>
      ))}
    </ul>
  )
}

export default () => (
  <Connect
    location="myorg.aragonid.eth"
    connector="thegraph"
    appConnectors={[

      // Everything app that matches would use the passed connector.
      // These rules would be applied one after another, so that we could
      // for example match all the voting apps with one connector, except one.
      [{ appName: 'voting.aragonpm.eth' }, VotingConnector],
      [{ appName: 'token-manager.aragonpm.eth' }, TokenConnector],

      // Like the .app() and .apps() filters
      [{ address: '0xcafe…' }, MyConnector],

      // It would also accept a function
      [app => app.appName = 'something.aragonpm.eth', MyOtherConnector],
    ]}
  >
    <App />
  </Connect>
)
```

packages/connect-react/src/index.ts
```diff
@@ -1,7 +1,139 @@
...
+  ConnectorDeclaration,
+  Organization,
+  Permission,
+  connect,
+} from '@aragon/connect'
+
+type ContextValue = {
+  org: Organization | null
+  orgError: Error | null
+  orgLoading: boolean
```
Evalir, line 14: Wondering if we should update the `org` types to `Organization` as discussed offline regarding naming?

packages/connect-react/src/index.ts
```diff
@@ -1,7 +1,139 @@
...
+    }
+  }, [location, connector, options])
+
+  const value = useMemo<ContextValue>(() => ({ org, orgError, orgLoading }), [
+    org,
+    orgError,
+    orgLoading,
+  ])
+
+  return React.createElement(ConnectContext.Provider, { value, children })
```
Evalir, line 80: Ooh interesting, why are we using `createElement` here? (Also, should we destructure it as we do with all the hooks?)

packages/connect-react/src/index.ts
```diff
@@ -1,7 +1,139 @@
...
+
+      handler = cbRef.current(org, (data: Data) => {
+        setData(data)
+        setLoading(false)
+      })
+    }
+    update()
+
+    return () => {
+      handler?.unsubscribe?.()
```
Evalir, line 118: ✨

packages/connect-react/src/index.ts
```diff
@@ -1,7 +1,139 @@
...
+    if (!org) {
+      return
+    }
+
+    let handler: any
+
+    const update = async () => {
+      setLoading(true)
+
+      handler = cbRef.current(org, (data: Data) => {
```
Evalir, line 110: I'm not quite getting what we're doing with the `org` parameter here. What's the thought process here?

packages/connect-react/src/index.ts
```diff
@@ -1,7 +1,139 @@
...
+  )
+}
+
+export function usePermissions(): PermissionsHookResult {
+  return useConnectSubscription<Permission[]>(
+    (org, onData) => org.onPermissions(onData),
+    []
+  )
+}
+
```
0xGabi, line 138: I wonder if we should have a hook in the lines of `useIntent`. This was something that 1Hive used when working on the honey-pot MVP. You can take a look at the specific logic of how they handle this: [here](https://github.com/1Hive/honey-pot/blob/master/src/hooks/useActions.js#L92).

packages/connect-thegraph/src/connector.ts
```diff
@@ -87,7 +87,7 @@ export default class ConnectorTheGraph extends GraphQLWrapper
     callback: Function
   ): { unsubscribe: Function } {
     return this.subscribeToQueryWithParser(
-      queries.ORGANIZATION_APPS('query'),
+      queries.ORGANIZATION_APPS('subscription'),
```
0xGabi, line 90: Nice catch! 💪

packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -62,7 +62,11 @@ export default class GraphQLWrapper {
 
         if (result.error) {
           throw new Error(
-            `Error performing subscription.${this.describeQueryResult(result)}`
+            [
```
0xGabi, line 65: Niceee

packages/connect/src/connect.ts
```diff
@@ -13,18 +13,22 @@ import ConnectorTheGraph, {
...
 import { Network } from '@aragon/connect-types'
 
-type ConnectOptions = {
+export type ConnectOptionsResolveIpfs = (
+  ipfsIdentifier: string,
+  path: string
+) => string
+
+export type ConnectOptions = {
   readProvider?: ethers.providers.Provider
```
0xGabi, line 22: I want to double-check if we still want to call this `readProvider` or just `provider`. Even though we only use it to read from the blockchain.

bpierre: Yes I don’t like `readProvider` much, let’s use `provider` (or `ethereum`?) and use the documentation to communicate that it’s read only.

packages/connect-react/src/index.ts
bpierre: I was thinking that we could keep using `org` internally and for variables in examples since it’s a very long word, and keep “organization” for anything that gets exposed (methods, types). What do you think?

packages/connect-react/src/index.ts
bpierre: > Ooh interesting, why are we using `createElement` here?

Just to avoid using JSX :-) This is the only element being created in this project, so I don’t think we really need JSX.

> (Also, should we destructure it as we do with all the hooks?)

I’m not sure I understand that one, could you please add a code example?

Evalir: My bad; I meant destructure `createElement`. :) As in

`import React, { createElement, ...} from 'react'`. Just for style!

packages/connect-react/src/index.ts
bpierre: The `useRef()` is only here to prevent `cb` to be added to the list of deps in the future (which would require to pass a memoized function): https://github.com/aragon/connect/blob/a3bc463f4f24853dad40b094ed67efce0423e874/packages/connect-react/src/index.ts#L120

Then we call `cb` with `org` and a `onData` callback, so that the final hooks (`useApps()` and `usePermissions()`) can initialize the subscription and return the handler. There is no need to memoize it never updates (thanks to the ref above): https://github.com/aragon/connect/blob/a3bc463f4f24853dad40b094ed67efce0423e874/packages/connect-react/src/index.ts#L127

examples/org-viewer-react/src/Group.tsx
```diff
@@ -0,0 +1,54 @@
+/** @jsx jsx */
+import { css, jsx, keyframes } from '@emotion/core'
+
+type GroupProps = {
+  children: React.ReactNode
+  loading?: boolean
+  name: string
+}
+
+// const blink = keyframes`50% { opacity: 0 }`
```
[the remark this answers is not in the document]
bpierre: Oops, this is an effect I was trying at some point for the loading states haha. Will remove :+1:

packages/connect-react/src/index.ts
bpierre: Oooh ok, yes let’s do this!

packages/connect-react/src/index.ts
bpierre: As discussed offline: let’s use `appIntent()` in the examples first, and see if an opportunity to provide a hook appear.

packages/connect-react/src/index.ts
Evalir: That makes sense. :) We just have to make sure to be very consistent with this separation though!

andy-hook:
We spoke about this offline so just to codify my thoughts on the error and loading status handling. I think having two items in the initial return might be a bit nicer:

`const [org, orgStatus] = useOrganization()`

And accessing properties:

`const { loading: boolean, error: null | Error, retry: Function } = orgStatus`

Or perhaps:

`const [state: { loading: boolean, error: null | Error}, retry: Function] = orgStatus`

I don't know if you had any further thoughts @bpierre

packages/connect/src/connect.ts
bpierre: Added it here https://github.com/aragon/connect/issues/109

packages/connect-react/src/index.ts
```diff
@@ -1,41 +1,46 @@
...
-type AppsHookResult = [Application[], Error | null, boolean]
-type PermissionsHookResult = [Permission[], Error | null, boolean]
+type ContextValue = {
+  org: Organization | null
+  orgLoadingStatus: LoadingStatus
+}
+type LoadingStatus = {
+  error: Error | null
+  loading: boolean
+  retry: () => void
```
bpierre, line 32: I think this is fine for this version, but since we are dealing with subscriptions, we should probably auto-retry for the next one.

- `error` would still get populated (and emptied on success).
- It would still be possible to call `retry`, e.g. we could imagine a user wanting to retry immediately by clicking a button.
- We could have a `retryEvery` prop on the provider, that either accept a number (milliseconds), or a function implementing a delay based on the number of attempts. Setting it to `-1` would disable the automatic retry behavior.

examples/org-viewer-react/README.md
```diff
@@ -0,0 +1,10 @@
+# Aragon Connect react examples
+
+This folder contains a small Aragon organization viewer built with react and Aragon Connect
```
sohkai, line 3:
```suggestion
This folder contains a small Aragon organization viewer built with React and Aragon Connect
```

examples/org-viewer-react/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+## How to run
+
+
+```
```
sohkai, line 8:
```suggestion

```
```

File: scripts/rebuild-packages.sh [hunk 13]
```diff
@@ -45,7 +45,7 @@ oao run-script --ignore-src '!packages/connect' build --parallel
 
 # Build the app connectors.
 echo 'Building the app connectors…'
-oao run-script --ignore-src '!packages/connect-thegraph-*' build --parallel
+oao run-script --ignore-src '!packages/connect-{thegraph-*,react}' build --parallel
```
sohkai, line 48: Should we be building the react library in this script too?

examples/org-viewer-react/webpack.config.js
```diff
@@ -0,0 +1,44 @@
+const path = require('path')
+const HtmlWebpackPlugin = require('html-webpack-plugin')
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
+  .BundleAnalyzerPlugin
```
sohkai, line 4: I just have to say this is such a weird syntax for `require` haha. I guess prettier did this?

examples/org-viewer-react/webpack.config.js
```diff
@@ -0,0 +1,44 @@
...
+  output: {
+    filename: 'bundle.js',
+    path: path.resolve(__dirname, 'dist'),
+  },
+  devServer: {
+    contentBase: path.join(__dirname, 'dist'),
+    port: 1234,
+  },
+  plugins: [
+    ...(process.env.BUNDLE_STATS === '1' ? [new BundleAnalyzerPlugin()] : []),
```
sohkai, line 41: Btw, how are we doing with this? What's this example clocking in at?

examples/org-viewer-react/src/use-routing.tsx
```diff
@@ -0,0 +1,29 @@
+import { Fragment, useEffect, useState } from 'react'
```
sohkai, line 1:
```suggestion
import { useEffect, useState } from 'react'
```

examples/org-viewer-react/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "scripts": {
+    "dev": "yarn clean && webpack-dev-server",
+    "build": "yarn clean && NODE_ENV=production webpack",
+    "build:stats": "yarn clean && webpack --env production --json > webpack-stats.json",
+    "clean": "rm -rf ./dist ./tsconfig.tsbuildinfo",
+    "compile": "tsc"
+  },
+  "dependencies": {
+    "@aragon/connect-react": "*",
+    "@emotion/core": "^10.0.28",
```
sohkai, line 16: How is `emotion` btw? Is it smaller than `styled-components`?

examples/org-viewer-react/src/OrgApps.tsx
```diff
@@ -0,0 +1,31 @@
...
+export default function OrgApps({ onOpenApp }: Props) {
+  const [apps, { loading }] = useApps()
+
+  return (
+    <Group name="Apps" loading={loading}>
+      <Table
+        headers={['Name', 'Version', 'Address']}
+        rows={[...apps]
+          .sort((a, b) => (a.name || '').localeCompare(b.name || ''))
+          .map(app => [
```
sohkai, line 21: Not a big deal, but I think you could make this a bit smaller by doing:

```jsx
rows={apps
  .map()
  .sort()
}
```

examples/org-viewer-react/src/OrgApps.tsx
```diff
@@ -0,0 +1,31 @@
...
+            app.name || 'unknown',
+            app.version || '?',
+            <TextButton onClick={() => onOpenApp(app.address)}>
+              {app.address.slice(0, 6)}
```
sohkai, line 25: I saw we're doing this in a few places, but should we use the `0xabcd....1234` format instead?

Know it's a demo, but still better to espouse good practices :).

examples/org-viewer-react/src/OrgGroups.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useOrganization } from '@aragon/connect-react'
+import OrgApps from './OrgApps'
+import OrgInfo from './OrgInfo'
+import OrgPermissions from './OrgPermissions'
+
+type OrgGroupsProps = { onOpenApp: (appAddress: string) => void }
+
+export default function OrgGroups({ onOpenApp }: OrgGroupsProps) {
+  const [org, { loading }] = useOrganization()
+  return (
```
sohkai, line 13: It's not a big deal here, but do we have an explicit contract that `org` will be unavailable until `loading` is false?

Or can you have a partial `org` object that is continually streamed in?

examples/org-viewer-react/src/OrgInfo.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export default function OrgApps() {
+  const [org] = useOrganization()
+
+  if (!org) {
+    return null
+  }
+
+  return (
+    <Group name="Organization">
```
sohkai, line 15: A few of the other components uses `loading` with `Group`—just wanted to double check we didn't want to do that here!

examples/org-viewer-react/src/OrgInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <Fragment>
+      <label>
+        <div
+          css={css`
+            padding-left: 4px;
+            padding-bottom: 8px;
+            font-size: 20px;
+          `}
+        >
+          Enter an org location:
```
sohkai, line 27:
```suggestion
          Enter an organization's location:
```

packages/connect-react/src/index.ts
sohkai: > Since we are dealing with subscriptions

Outside of the `retryEvery` prop or a user calling `retry`, the subscription would only help here if the underlying data was updated in the mean time, right? Or could you see a case where you hadn't connected / had some sort of corruption, and the GraphQL subscription just magically fixed it at a later point?

packages/connect-react/src/index.ts
```diff
@@ -1,7 +1,162 @@
...
+  ConnectorDeclaration,
+  Organization,
+  Permission,
+  connect,
+} from '@aragon/connect'
+
+type ConnectProps = {
+  children: React.ReactNode
+  connector: ConnectorDeclaration
+  location: string
```
sohkai, line 22: `location` feels a bit vague, especially as a keyed React prop. It works a lot better as a positional argument's name and I don't have any better suggestions!

packages/connect-react/src/index.ts
```diff
@@ -1,7 +1,162 @@
...
+  cb: (
+    org: Organization,
+    onData: (data: Data) => void
+  ) => { unsubscribe: Function },
+  initValue: Data
+): [Data, LoadingStatus] {
+  const [data, setData] = useState<Data>(initValue)
+  const [loading, setLoading] = useState<boolean>(false)
+  const { org } = useContext<ContextValue>(ConnectContext)
+  const cbRef = useRef(cb)
```
sohkai, line 119: I think this was discussed earlier (https://github.com/aragon/connect/pull/52#discussion_r448598635), but the API feels kind of awkward to enforce that "your callback may never change".

Otherwise, it might be worth adding a comment to explain that this hook expects its parameters to never change.

packages/connect-react/README.md
```diff
@@ -0,0 +1,64 @@
+# Aragon Connect for React
+
+## Introduction
+
+Aragon Connect for React provides a series of utilities that make Aragon Connect integrated into a React environment.
```
sohkai, line 5:
```suggestion
Aragon Connect for React provides a series of utilities that simplify the usage of Aragon Connect in a React environment.
```

packages/connect-react/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+It consists of the `<Connect />` component, on which a connection to an organization is described, and a series of hooks: `useApps()`, `useOrganization()`, `usePermissions()`.
```
sohkai, line 7:
```suggestion
It consists of the `<Connect />` component, through which a connection to an organization is described, and a series of hooks: `useApps()`, `useOrganization()`, `usePermissions()`.
```

packages/connect-react/README.md
```diff
@@ -0,0 +1,64 @@
...
+```jsx
+import {
+  Connect,
+  useApps,
+  useOrganization,
+  usePermissions,
+} from '@aragon/connect-react'
+
+function App() {
+  // connect-react functions return three things: the data, the error, and the loading state.
```
sohkai, line 20:
```suggestion
  // connect-react hooks return three things: the data, the error, and the loading state.
```

But I think this needs to be changed, right? Should we point to the types?

packages/connect-react/README.md
```diff
@@ -0,0 +1,64 @@
+# Aragon Connect for React
```
sohkai, line 1: Should we also leave some notes in the gitbooks docs for this library?

scripts/rebuild-packages.sh
sohkai: Oh nevermind, I see the default build part will build the react library too!

examples/org-viewer-react/package.json
bpierre: > How is `emotion` btw?

It works fine, not that different from `styled-components` so far. I haven’t had a close look at the generated CSS code yet.

One nice thing is that we can use it with the `@jsx` pragma, to get a `css` prop without a babel plugin. It is a bit more verbose than the styled babel plugin (`css={css``}` + the import + the pragma directive), but I find it acceptable.

> Is it smaller than `styled-components`?

Yes, 6kb vs. 12kb:

- https://bundlephobia.com/result?p=@emotion/core@10.0.28
- https://bundlephobia.com/result?p=styled-components@5.1.1

examples/org-viewer-react/src/OrgApps.tsx
bpierre: Yes good idea!

packages/connect-react/src/index.ts
bpierre: What it be `organization`? I avoided it because we tend to use it for `Organization` instances, but I don’t think that’s a big issue.

In other places, I am trying to use `location` when both an Ethereum address and an ENS domain are accepted, and `address` when only an Ethereum address is accepted.

packages/connect-react/src/index.ts
bpierre: Yes you’re right, we might want to change that. Especially now that we have parameters on `useApps()` and `useApp()`.

examples/org-viewer-react/src/OrgGroups.tsx
bpierre: At the moment, `org` is `null` until it finishes loading, but it only does it once. Ideally I think we should keep streaming continually yes (that would require a new method on the connector).

packages/connect-react/src/index.ts
0xGabi: I quite like `location` I think in the context of the Connect component should not be that vague. Otherwise how you feel about `identity`?

examples/org-viewer-react/webpack.config.js
bpierre: 1.01 MB for this demo in total, 59.23% of which is ethers.js.

image

examples/org-viewer-react/src/OrgInfo.tsx
bpierre: I refactored the loading states a bit, so that every box is doing the same thing now.

packages/connect-react/src/index.ts
bpierre: Done!

examples/org-viewer-react/src/OrgPermissions.tsx
```diff
@@ -0,0 +1,25 @@
+/** @jsx jsx */
+import { jsx } from '@emotion/core'
+import { usePermissions } from '@aragon/connect-react'
+import Group from './Group'
+import Table from './Table'
+
+export default function OrgPermissions() {
+  const [permissions, { loading }] = usePermissions()
+  return (
+    <Group name="Permissions" loading={loading}>
```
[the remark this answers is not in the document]
bpierre: Yes: if we are waiting for the first `onPermissions()` call, or if the org itself is loading.

packages/connect-react/src/index.ts
bpierre: `identity` feels a bit too far from the idea of “location of your organization” IMO, I would prefer to keep `location` until we find something better.

For now I think any of these would work:

- `address` (“address of the organization”, we could clarify in the docs that it accepts ENS domains)
- `organization` (we could specify in the docs that it accepts a string and not an `Organization`).
- `location`

sohkai: `location` sounds good—let's just make sure the concept of an organization's "location" (address or ENS domain resolving to a valid address) is in the docs!

PR #67, opened 2020-06-20
Add packages.json files config

0xGabi:


packages/connect-core/package.json
```diff
@@ -15,6 +15,10 @@
     "compile": "tsc --build",
     "lint": "eslint --ext .ts ./src"
   },
+  "files": [
+    "dist",
+    "src"
```
bpierre, line 20: I think we could exclude `src/`, since even the types are in `dist/`?

PR #68, opened 2020-06-21
Add support for xdai network

0xGabi:
1Hive is deploying Aragon Infrastructure to xDai and this is a requirement for `connect` to keep compatible with it.

### DIFF
--- a/packages/connect-thegraph/subgraph/manifest/data/xdai.json
+++ b/packages/connect-thegraph/subgraph/manifest/data/xdai.json
@@ -1,0 +1,22 @@
+{
+  "network": "xdai",
+  "description": "Development subgraph for Aragon organizations.",
+  "repository": "https://github.com/aragon/connect",
+  "ens": "0xaafca6b0c89521752e559650206d7c925fd0e530",
+  "DAOFactoryDataSources": [
+    {
+      "name": "DAOFactory",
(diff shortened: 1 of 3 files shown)


packages/connect/src/connect.ts
```diff
@@ -77,6 +77,13 @@ function getNetwork(chainId?: number): Network | null {
       ensAddress: '0x98df287b6c145399aaa709692c8d308357bc085d',
     }
   }
+  if (chainId === 100) {
+    return {
+      chainId: 100,
+      name: 'xdai',
```
bpierre, line 83: Just as a note, I was thinking of starting to accept the chain as `{ chainId: number, name: string }` rather than only `chainId: number`, to handle staging environments.

packages/connect-thegraph/subgraph/manifest/data/xdai.json
```diff
@@ -0,0 +1,22 @@
+{
+  "network": "xdai",
+  "description": "Development subgraph for Aragon organizations.",
```
sohkai, line 3:
```suggestion
  "description": "Subgraph for Aragon organizations.",
```

packages/connect/src/connect.ts
0xGabi: Nice, I like that.

PR #73, opened 2020-06-24
Refactor transactions

0xGabi:
Update the `Transaction Request` object to keep it simple:

```
{
  data: string
  from: string
  to: string
}
```

Add a new transaction entity, `Transaction Described`. Has the list of transction with their Radscpec description.

Refactor the transaction path logic to remove the gas estimation. We discuss this behaviour is more align with the Principles of the Connect library and push that responsibility to the wallet.


api-reference/transaction-described.md
```diff
@@ -0,0 +1,13 @@
+# TransactionRequest
```
bpierre, line 1:
```suggestion
# TransactionDescribed
```

api-reference/transaction-intent.md
```diff
@@ -8,25 +8,24 @@ An TransactionIntent to create an Ethereum transaction on, or through, an Aragon
...
-| returns | `TransactionPath[]` | Array of all the possible transaction paths. |
+| Name           | Type                         | Description                                                                                               |
+| :------------- | :--------------------------- | :-------------------------------------------------------------------------------------------------------- |
+| `account`      | `String`                     | The account that will sign the transaction.                                                               |
+| `options`      | `Object`                     | Options object.                                                                                           |
+| `options.as`   | `String`                     | Address of an Aragon organization, or its agent app, through which the paths should get created.          |
+| `options.path` | `String[]`                   | An array of address that conform a transaction path, it will be verified without calculating other paths. |
+| returns        | `Promise<TransactionPath[]>` | Array of all the possible Transaction Paths.                                                              |
```
bpierre, line 17: For clarity, I think we should either use a descriptive name (“transaction paths”) or the exact class name (`TransactionPath`) in the documentation.

```suggestion
| returns        | `Promise<TransactionPath[]>` | Array of all the possible transaction paths.                                                              |
```

api-reference/transaction-path.md
```diff
@@ -4,9 +4,10 @@ Represents a single transaction path.
...
-| Name | Type | Description |
-| :--- | :--- | :--- |
-| `apps` | `App[]` | Get all the apps for the path. |
-| `destination` | `App` | Get the destination of the transactions path. |
-| `transactions` | `TransactionRequest[]` | Transaction request objects, ready to be sent to an Ethereum library. |
-
+| Name                          | Type                     | Description                                                                    |
+| :---------------------------- | :----------------------- | :----------------------------------------------------------------------------- |
+| `apps`                        | `App[]`                  | Get all the apps for the path.                                                 |
+| `transactionsDescribed`       | `TransactionDescribed[]` | List of transactions in the path with their corresponding radspec description. |
```
bpierre, line 10: Would it make more sense to move this on `TransactionRequest#description` and `TransactionRequest#descriptionAnnotated`? I don’t think it would be a problem for any Ethereum library, but we could also add a `.export()` that exports a clean transaction request object (only `to` / `from` / `data`), in case a specific context needs it.

examples/nodejs/package.json
```diff
@@ -23,7 +23,7 @@
     "@aragon/connect-thegraph": "*",
     "@aragon/connect-thegraph-token-manager": "*",
     "@aragon/connect-thegraph-voting": "*",
-    "ethers": "^5.0.0-beta.188",
+    "ethers": "^5.0.0",
```
bpierre, line 26: :partying_face:

packages/connect-core/src/utils/transactions.ts
```diff
@@ -173,10 +108,57 @@ export function createForwarderTransactionBuilder(
...
+    throw new Error(
+      `Balance too low. ${from} balance of ${tokenAddress} token is ${balance} (attempting to send ${tokenValue})`
+    )
+  }
+
+  const allowance = await tokenContract.allowance(from, approveSpender)
+  const allowanceBN = BigInt(allowance)
+  // If allowance is already greater than or equal to amount, there is no need to do an approve transaction
+  if (allowanceBN < tokenValueBN) {
+    if (allowanceBN > BigInt(0)) {
```
bpierre, line 139:
```suggestion
    if (allowanceBN > 0n) {
```

packages/connect-core/src/transactions/TransactionRequest.ts
```diff
@@ -21,22 +12,17 @@ export interface Annotation {
...
-  readonly gasLimit?: string
-  readonly chainId?: number
+  readonly description?: string
+  readonly data!: string
+  readonly from!: string
+  readonly to!: string
 
   constructor(data: TransactionRequestData) {
-    Object.assign(this, data)
+    this.annotatedDescription = data.annotatedDescription
```
bpierre, line 22: Should we use `description` and `descriptionAnnotated`, like we have `transactions` and `transactionDescribed`? That way they will both complete when `description…` is typed, and it makes them a bit easier to identify as being related.

0xGabi: Yeah, good point

api-reference/transaction-request.md
```diff
@@ -4,14 +4,10 @@ An object describing a transaction that can get signed by a library like ethers.
...
-| `from` | `String` | Address to use as default sender. |
-| `gas` | `String` | Price \(in wei\) per unit of gas. Duplicate of `gasPrice`. |
-| `gasLimit` | `Number` | Maximum gas this transaction may spend. |
-| `gasPrice` | `String` | Price \(in wei\) per unit of gas. |
-| `to` | `String` | Target address or ENS name. |
-| `value` | `String` | Amount \(in wei\) this transaction is sending. |
-
+| Name                   | Type           | Description                                                                                  |
+| :--------------------- | :------------- | :------------------------------------------------------------------------------------------- |
+| `annotatedDescription` | `Annotation[]` | List of the Radspec description bindings with the properties `{ type: string, value: any }`. |
```
bpierre, line 9:
```suggestion
| `descriptionAnnotated` | `Annotation[]` | List of the Radspec description bindings with the properties `{ type: string, value: any }`. |
```

PR #78, opened 2020-06-26
Revert "GitBook: [master] 14 pages modified"

sohkai:
This reverts commit 77f6048699bbac586778d33f11c303784c4a0364.

Not sure how Gitbooks is managing the sync, but I'm pretty sure the version on our side on Github was newer.

I've updated Gitbooks so it only syncs from master (we can use `v*` branches in the future to save version variants).


bpierre:
> Not sure how Gitbooks is managing the sync, but I'm pretty sure the version on our side on Github was newer.

Yes, not sure why GitBook keeps doing that. @0xGabi any idea?

> I've updated Gitbooks so it only syncs from master (we can use `v*` branches in the future to save version variants).

Does it means that it will only get updates from `master`, and stop pushing to it?

sohkai:
> Yes, not sure why GitBook keeps doing that

So this has happened because I created/changed variants on GitBook's side, and had to push the change through their side vs. on Github. It doesn't appear as though there's a way to configure the variants via a configuration file, outside of just using different branches.

> Does it means that it will only get updates from master, and stop pushing to it?

It will still be able to push to it if we edit anything from the GUI, but for now I think we should stop doing that :)

PR #89, opened 2020-06-29
Feat: Add IPFS fallback logic to fetch metadata

0xGabi:
This PR includes a new async method `create()` for the entities that need metadata (App, Repo, and Role). This a requirement when creating the entity. We handle this during creation at the connector.

If we can't get the metadata from The Graph we use ethers to fetch with a GET request constructing the URL from the `contentUri` pointing to the default Aragon IPFS node.


packages/connect-core/src/entities/Organization.ts
```diff
@@ -86,7 +86,11 @@ export default class Organization {
...
     this.checkConnected()
-    return this._connector.appsForOrg(this.address)
+    const apps = await this._connector.appsForOrg(this.address)
+    for (const app of apps) {
+      await app._init()
+    }
+    return apps
   }
 
   onApps(callback: Function): { unsubscribe: Function } {
```
0xGabi, line 96: I'm not sure how to handle the `_init()` method for subscriptions. Ideas are welcome 😁

packages/connect-core/src/entities/Organization.ts
```diff
@@ -86,7 +86,11 @@ export default class Organization {
   ///////// APPS ///////////
   async apps(): Promise<Application[]> {
     this.checkConnected()
-    return this._connector.appsForOrg(this.address)
+    const apps = await this._connector.appsForOrg(this.address)
+    for (const app of apps) {
+      await app._init()
```
0xGabi, line 91: I wonder what you think of handling here like this. I was not sure if a `Promise.all` was a better pattern in this case.

Evalir: I think it may be better to handle them using `Promise.all` as they will be able to be resolved in parallel rather than sequentially. 😃

packages/connect-core/src/entities/Repository.ts
```diff
@@ -20,60 +20,63 @@ export interface RepositoryData {
...
-      this.changelogUrl = changelogUrl
-      this.description = description
-      this.descriptionUrl = descriptionUrl
-      this.icons = icons
-      this.screenshots = screenshots
-      this.sourceUrl = sourceUrl
-    }
-
     this.address = data.address
+    this.contentUri = data.contentUri ?? undefined
```
Evalir, line 43: ES2020 🤩 wouldn't it be better to return `null`?

packages/connect-core/src/entities/Role.ts
```diff
@@ -1,45 +1,52 @@
...
-
     this.appAddress = data.appAddress
+    this.hash = data.hash
     this.permissions = data.grantees?.map(
       grantee => new Permission(grantee, connector)
     )
-    this.hash = data.hash
     this.manager = data.manager
+    this.#artifact = data.artifact
+    this.#contentUri = data.contentUri ?? undefined
```
Evalir, line 37: Same as above; should we return `null`?

packages/connect-core/src/entities/Repository.ts
0xGabi: Thank for the suggestions. I think in this particular case the problem is that we are implementing the `RepositoryData` interface and the `contentUri` has those types. Regardless I agree there is something a bit odd here, and I believe it's a consequence of allowing data on The Graph to be `undefined / null`. 

Anyways I like to know why you think `null` would be better than `undefine`?

Evalir: It's probably due to my mental model regarding `null` & `undefined`; for me, `null` is the proper thing to use when there's _no_ value. It indicates the absence of it. `undefined`, for me, indicates problems with fetching some data, a function that returns nothing (`void`) or some other anomalous development in processing data.

packages/connect-core/src/entities/Application.ts
```diff
@@ -24,71 +24,35 @@ export interface ApplicationData {
...
+  abi?: Abi
+  appName?: string
+  author?: string
+  contractPath?: string
+  deprecatedIntents?: { [version: string]: AppIntent[] }
+  description?: string
+  icons?: { src: string; sizes: string }[]
+  intents?: AppIntent[]
+  htmlUrl?: string
+  sourceUrl?: string
```
bpierre, line 51: Ah I didn’t think about that… I’m wondering if it’s worth using `readonly` at all then?

packages/connect-core/src/entities/Application.ts
```diff
@@ -101,13 +65,61 @@ export default class Application extends CoreEntity implements ApplicationData {
...
+      author,
+      description,
+      start_url: htmlUrl,
+      icons,
+      source_url: sourceUrl,
+    }: AragonManifest = await resolveMetadata(
+      'manifest.json',
+      this.contentUri!,
+      this.#manifest
+    )
```
bpierre, line 95: We should probably try / catch these two, and emit contextual errors.

packages/connect-core/src/entities/Role.ts
bpierre: I’m thinking that the only reason to return `undefined` here would be for `contentUri` to be `undefined`, since `??` would exclude an empty string. Should we use `||` instead?

packages/connect-core/src/entities/Role.ts
```diff
@@ -1,45 +1,52 @@
...
+  }
+
+  async _init(): Promise<void> {
+    const { roles }: AragonArtifact = await resolveMetadata(
+      'artifact.json',
+      this.#contentUri!,
+      this.#artifact
+    )
```
bpierre, line 45: Same here, we should probably try / catch.

packages/connect-core/src/entities/Repository.ts
bpierre: Agreed on the mental model, I see `null` as “empty”, while `undefined` is “nothing”. I also like to keep using the same type when it makes sense, e.g. here we could use `""`, or if it was a list we could use `[]`: it removes the need to treat `null` / `undefined` as an exception for the consumer.

…I wish JS only had `null` :laughing:

packages/connect-core/src/entities/Application.ts
0xGabi: If we are going to use a method `_init()` / `create()` always required probably not worth it. I can't think other alternatives to try to have `readonly` otherwise.

packages/connect-core/src/entities/Role.ts
Evalir: Gotcha; yes this makes sense to me.

packages/connect-core/src/entities/Application.ts
0xGabi: I decide to handle it in the `resolveMetadata` function. What you think?

packages/connect-core/src/entities/Repository.ts
0xGabi: Now that you mention about using an empty object of the particular type. I was wondering if we should do that at the subgraph level and always asure that at least we have an empty object. At least for the most part of the attributes.

0xGabi:
As we discussed with Pierre I move the logic of `_init()` to `create()` instead. Then I move the method call to the connector so we have both the instantiation of a new object and the call to `create()` method together.

0xGabi:
Note I split the PRs to keep the changes more clear.

sohkai:
> We use ethers to fetch with a GET request constructing the URL from the contentUri pointing to the default Aragon IPFS node

I was thinking about this (moreso in the context of a CLI/frontend using connect to connect to generic apps that may or may not be hosted by our default IPFS node), and I think it's important to provide a way to override the default IPFS gateway used.

packages/connect-core/src/entities/Repository.ts
bpierre: > Now that you mention about using an empty object of the particular type. I was wondering if we should do that at the subgraph level and always asure that at least we have an empty object. At least for the most part of the attributes.

I think it could be nice yes. :+1:

packages/connect-core/src/entities/App.ts
```diff
@@ -26,81 +26,91 @@ export interface AppData {
...
 
-    // TODO: If no metadata, fallback to resolve ourselves with ipfs
+    this.#created = false
+  }
+
+  get created(): boolean {
+    return this.#created
+  }
+
+  async create({ artifact, manifest, ...data }: AppData): Promise<void> {
```
bpierre, line 63: For `create`, I was more thinking of a static method, that would fetch everything needed and instantiate, so we can do this:

```jsx
const app = await App.create(/* … */)
```

Rather than this:

```jsx
const app = new App(/* … */)
await app.create()
```

That way, `App` would never be in a state where it is instantiated, but not initialized.

packages/connect-core/src/entities/App.ts
```diff
@@ -26,81 +26,91 @@ export interface AppData {
...
+        manifest
+      )
 
+      this.abi = abi
+      this.address = data.address
+      this.appId = data.appId
+      this.appName = appName
       this.author = author
+      this.codeAddress = data.codeAddress
+      this.contentUri = data.contentUri || undefined
```
bpierre, line 97: I think we can do this?

```suggestion
      this.contentUri = data.contentUri
```

packages/connect-core/src/entities/App.ts
```diff
@@ -26,81 +26,91 @@ export interface AppData {
...
       this.description = description
-      this.htmlUrl = htmlUrl
       this.icons = icons
+      this.intents = functions
+      this.isForwarder = data.isForwarder ?? undefined
+      this.isUpgradeable = data.isUpgradeable ?? undefined
+      this.htmlUrl = htmlUrl
+      this.kernelAddress = data.kernelAddress
+      this.name = data.name
+      this.registry = data.registry || undefined
```
bpierre, line 108:
```suggestion
      this.registry = data.registry
```

packages/connect-core/src/entities/App.ts
0xGabi: Oh, I didn't think about that. I'll include this change as a new PR.

PR #92, opened 2020-06-30
Docs: Small fixes & correct links

0xGabi:


docs/SUMMARY.md
```diff
@@ -1,25 +1,25 @@
 # Table of contents
 
-* [Aragon Connect](README.md)
+- [Aragon Connect](README.md)
```
bpierre, line 3: I think we should keep `*`, otherwise GitBook will override our branch.

docs/api-reference/transactionintent.md
```diff
@@ -1,6 +1,6 @@
 # Intent
```
bpierre, line 1:
```suggestion
# TransactionIntent
```

docs/api-reference/transactionintent.md
```diff
@@ -1,6 +1,6 @@
...
 
-An TransactionIntent to create an Ethereum transaction on, or through, an Aragon organization or its apps. It allows to retrieve transaction paths or a set of transactions to sign.
+An `TransactionIntent` to create an Ethereum transaction on, or through, an Aragon organization or its apps. It allows to retrieve transaction paths or a set of transactions to sign.
```
bpierre, line 3: Should we use descriptive words here?

```suggestion
An intent to create an Ethereum transaction on, or through, an Aragon organization or its apps. It allows to retrieve transaction paths or a set of transactions to sign.
```

PR #101, opened 2020-07-03
Feat: Add descriptions utils for evm scripts and transactions

0xGabi:
This PR includes a utility to describe EVM scripts and transactions. 

It exposes a few of the utils we use on `connect-core` to be accessible from `connect`. 

It includes an example of how to use `decodeScript` to decode a script from a vote.


examples/nodejs/src/describe-script.ts
```diff
@@ -0,0 +1,40 @@
+import { connect, describeScript, App } from '@aragon/connect'
```
0xGabi, line 1: Example of how to use `describeScript` to decode script from votes.

packages/connect-core/src/utils/index.ts
```diff
@@ -0,0 +1,4 @@
+export { findAppMethodFromIntent } from './app'
```
0xGabi, line 1: I decided to expose this utilities, what you think?

packages/connect-core/src/utils/descriptions.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+export async function describeTransactionPath(
+  path: TransactionWithChildren[],
+  apps: App[],
+  provider?: ethers.providers.Provider
+): Promise<TransactionRequest[]> {
+  return Promise.all(
+    path.map(async step => {
+      return describeTransaction(step, apps, provider)
+    })
```
bpierre, line 66:
```suggestion
    path.map(step => describeTransaction(step, apps, provider))
```

packages/connect-core/src/utils/index.ts
bpierre: Sounds good yes!

PR #110, opened 2020-07-04
Refactor: Use an static method to create Entities + Split Data and Metadata

0xGabi:


docs/api-reference/app.md
```diff
@@ -10,22 +10,19 @@ An Aragon app installed in an organization from an [aragonPM repository](https:/
 | `address` | `String` | The address of the app proxy contract \(never changes\). |
 | `appId` | `String` | The `appName` encoded via [namehash](https://eips.ethereum.org/EIPS/eip-137). |
 | `appName` | `String` | The app's ENS identifier. E.g. `"token-manager.aragonpm.eth"` |
```
0xGabi, line 12: A few of the properties now are `get` methods. Should we keep them as properties or move them as methods?

packages/connect-core/src/entities/App.ts
```diff
@@ -25,92 +31,63 @@ export interface AppData {
...
+      data.artifact
+    )
+
+    const manifest: AragonManifest = await resolveMetadata(
+      'manifest.json',
+      data.contentUri || undefined,
+      data.manifest
+    )
+
+    const metadata: Metadata = [artifact, manifest]
```
0xGabi, line 88: Now we have a clear distinction between data and metadata. Probably during creation as well we will handle the overrides of the few contract that miss repos on aragonPM (Kernel, ACL, EVM, etc)

docs/api-reference/app.md
bpierre: I think it’s better to keep them as property, since that’s how they feel from the outside.

packages/connect-core/src/entities/App.ts
```diff
@@ -120,4 +97,28 @@ export default class App extends CoreEntity implements AppData {
...
+
+  get artifact(): AragonArtifact {
+    return this.#metadata[0] as AragonArtifact
+  }
+
+  get manifest(): AragonManifest {
+    return this.#metadata[1] as AragonManifest
+  }
+
+  get abi(): Abi {
```
0xGabi, line 109: I only include getter for really useful properties. Otherwise, the dev can get them from the metadata files.

packages/connect-thegraph/src/parsers/repos.ts
```diff
@@ -22,8 +22,5 @@ export async function parseRepo(
     registryAddress: app.repo?.registry?.address,
   }
 
-  const repoEntity = new Repo(connector)
-  await repoEntity.create(data)
-
-  return repoEntity
+  return Repo.create(data, connector)
```
0xGabi, line 25: This is how we create the entities now.

PR #111, opened 2020-07-05
App filters + align Organization methods

bpierre:
Fixes https://github.com/aragon/connect/issues/48

- Add an `Organization#onApp()` method.
- Add app filters to:
  - `Organization#app()`
  - `Organization#apps()`
  - `Organization#onApp()`
  - `Organization#onApps()`
- Update Organization API docs (remove anything not implemented).


docs/guides/getting-started.md
```diff
@@ -73,8 +73,13 @@ As seen above, connecting to an organization can be done by calling the `connect
 
 It requires two parameters:
 
+<<<<<<< HEAD
```
0xGabi, line 76: This file still has a few merge conflicts.

packages/connect-core/src/connections/ConnectorInterface.ts
```diff
@@ -2,14 +2,28 @@ import App from '../entities/App'
 import Permission from '../entities/Permission'
 import Repo from '../entities/Repo'
 import Role from '../entities/Role'
+import { AppFilters } from '@aragon/connect-types'
 
 export interface ConnectorInterface {
+  appByAddress(appAddress: string): Promise<App>
```
0xGabi, line 8: Now that we have the filter. Does the interface need to implement this method?

packages/connect-core/src/entities/Organization.ts
```diff
@@ -81,102 +116,63 @@ export default class Organization {
...
+    )
+  }
+
+  onApps(
+    filtersOrCallback: AppFiltersParam | OnAppsCallback,
+    callback?: OnAppsCallback
+  ): SubscriptionHandler {
+    this.checkConnected()
+
+    const filters = (callback ? filtersOrCallback : null) as AppFiltersParam
```
0xGabi, line 155: Why the `filters` are only applied when a callback exists?

packages/connect-thegraph/src/connector.ts
```diff
@@ -31,6 +31,22 @@ function getOrgSubgraphUrl(network: Network): string | null {
   return null
 }
 
+function appFiltersToQueryFilter(appFilters: AppFilters) {
+  const queryFilter = {} as any
+
+  if (appFilters.name) {
+    queryFilter.repoName_in = appFilters.name.map(name =>
```
0xGabi, line 38: Maybe we should also filter the registry? One idea is that `queryFilter` be a tuple: `[name, registry]`

docs/guides/getting-started.md
bpierre: oops :see_no_evil:

packages/connect-core/src/connections/ConnectorInterface.ts
bpierre: Yes this is still useful to retrieve an app from any organization, while the other one is only for the current organization.

packages/connect-core/src/entities/Organization.ts
bpierre: This is to make the first parameter optional, so that we could do this:

```tsx
org.onApps('voting', apps => {
})
```

And this:

```tsx
org.onApps(apps => {
})
```

packages/connect-thegraph/src/connector.ts
bpierre: Do you think that would be useful initially?

I’m thinking that it might be tricky to implement, as I couldn’t find a way to use a OR at the `where` level. This is why it is not possible right now to pass e.g. an address and a different app name.

If we also filter by registry, I think we would have the same kind of problem, where a list of names using different registries would be a problem as a query, e.g.

```tsx
const apps = await org.apps({
  appName: [
    ['voting', 'aragonpm.eth'],
    ['token-manager', 'somethingelse.eth']
  ]
})
```

I was thinking about having this as a base for now, and investigate together on the filters we want and maybe find a way to express an OR in the `where` expression.

packages/connect-core/src/connections/ConnectorInterface.ts
0xGabi: Right, thanks for clarifying.

packages/connect-core/src/entities/Organization.ts
0xGabi: Nice, now I understood the semantic of the syntax.

packages/connect-thegraph/src/connector.ts
0xGabi: Make sense, as we discuss async let's think more about it next week. I agree this is good enough for now.

PR #114, opened 2020-07-05
Feat: Add overrides for apps without aragonPM repos

0xGabi:
Includes manual override of the artifact metadata for apps without an aragonPM repo:
- Kernel
- ACL
- EVM Script Registry
- Repo
- APM Registry
- ENS Subdomain Registrar

### DIFF
--- a/packages/connect-core/src/entities/App.ts
+++ b/packages/connect-core/src/entities/App.ts
@@ -10,6 +10,11 @@ import {
 } from '../types'
 import { resolveMetadata } from '../utils/metadata'
 import { ConnectorInterface } from '../connections/ConnectorInterface'
+import {
+  getApmInternalAppInfo,
+  getAragonOsInternalAppInfo,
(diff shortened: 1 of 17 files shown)


packages/connect-core/src/entities/App.ts
```diff
@@ -73,11 +78,18 @@ export default class App extends CoreEntity {
...
+      artifact = getApmInternalAppInfo(data.appId)
+    } else if (hasAppInfo(data.appId, 'aragon')) {
+      artifact = getAragonOsInternalAppInfo(data.appId)
+    } else {
+      artifact = await resolveMetadata(
+        'artifact.json',
+        data.contentUri || undefined,
+        data.artifact
+      )
+    }
```
bpierre, line 92: I was wondering if we should move this in a utility function? (Probably something we can do later)

0xGabi: Yes, good idea!

PR #118, opened 2020-07-06
Docs: Use \| as separators

0xGabi:
@bpierre I wonder if maybe we should use `or` as we did in a few places.


bpierre:
> @bpierre I wonder if maybe we should use `or` as we did in a few places.

Yes let’s do this! We were talking with @sohkai about perhaps defining somewhere this pseudo language we use in the documentation (e.g. “How to read this documentation” in the intro page).

PR #121, opened 2020-07-07
Fix: solve metadata for Roles + TransactionRequest#children

0xGabi:
Add a few small fixes from user feedback.

### DIFF
--- a/packages/connect-core/src/entities/App.ts
+++ b/packages/connect-core/src/entities/App.ts
@@ -8,13 +8,8 @@ import {
   Abi,
   AppIntent,
 } from '../types'
-import { resolveMetadata } from '../utils/metadata'
+import { resolveManifest, resolveArtifact } from '../utils/metadata'
 import { ConnectorInterface } from '../connections/ConnectorInterface'
-import {
-  getApmInternalAppInfo,
-  getAragonOsInternalAppInfo,
-} from '../utils/overrides/index'
-import { hasAppInfo } from '../utils/overrides/interfaces'
 
 // TODO:
 // [ ] (ipfs) contentUrl 	String 	The HTTP URL of the app content. Uses the IPFS HTTP provider. E.g. http://gateway.ipfs.io/ipfs/QmdLEDDfi…/ (ContentUri passing through the resolver)
@@ -78,24 +73,8 @@ export default class App extends CoreEntity {
     data: AppData,
     connector: ConnectorInterface
   ): Promise<App> {
-    let artifact: AragonArtifact
-    if (hasAppInfo(data.appId, 'apm')) {
-      artifact = getApmInternalAppInfo(data.appId)
-    } else if (hasAppInfo(data.appId, 'aragon')) {
-      artifact = getAragonOsInternalAppInfo(data.appId)
-    } else {
-      artifact = await resolveMetadata(
-        'artifact.json',
-        data.contentUri || undefined,
-        data.artifact
-      )
-    }
-
-    const manifest: AragonManifest = await resolveMetadata(
-      'manifest.json',
-      data.contentUri || undefined,
-      data.manifest
-    )
+    const artifact = await resolveArtifact(data)
+    const manifest = await resolveManifest(data)
 
     const metadata: Metadata = [artifact, manifest]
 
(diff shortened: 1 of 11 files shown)


packages/connect-core/src/entities/Organization.ts
```diff
@@ -105,6 +105,11 @@ export default class Organization {
     return this.#address || '' // The || '' should never happen but TypeScript requires it.
   }
 
+  get provider(): ethers.providers.Provider {
```
0xGabi, line 108: May be useful to have access to the same provider the connect library initialize.

packages/connect-core/src/entities/Role.ts
```diff
@@ -31,7 +33,7 @@ export default class Role extends CoreEntity {
 
     const { roles } = metadata[0] as AragonArtifact
 
-    const role = roles.find(role => role.bytes === data.hash)
+    const role = roles?.find(role => role.bytes === data.hash)
```
0xGabi, line 36: We had an error here when the `roles` were undefined.

packages/connect-thegraph/src/parsers/roles.ts
```diff
@@ -33,8 +28,9 @@ async function _parseRole(
     manager: role.manager,
     hash: role.roleHash,
     grantees,
-    artifact,
-    contentUri,
+    appId: app.appId,
```
0xGabi, line 31: Includes the `appId` on the `Role` entity so we can override the artifact metadata for the apps without a repo on aragonPM.

packages/connect-core/src/entities/App.ts
```diff
@@ -78,24 +73,8 @@ export default class App extends CoreEntity {
...
-      )
-    }
-
-    const manifest: AragonManifest = await resolveMetadata(
-      'manifest.json',
-      data.contentUri || undefined,
-      data.manifest
-    )
+    const artifact = await resolveArtifact(data)
+    const manifest = await resolveManifest(data)
```
bpierre, line 77: ✨

packages/connect-core/src/entities/Organization.ts
bpierre: Good point yes!

I think we could do this for now, and when making the “connection context” changes, move this inside a single `.connection` property that would contain the provider, the connector, etc. That way, we wouldn’t add too many properties on `Organization` that are in fact related to the connection.

packages/connect-core/src/entities/Role.ts
bpierre: Ah yes and we are using `as` before so TS couldn’t catch it :s

packages/connect-core/src/utils/descriptions.ts
```diff
@@ -40,9 +40,9 @@ export async function describeTransaction(
         provider
       )
     }
-
-    // eslint-disable-next-line no-empty
-  } catch (_) {}
+  } catch (err) {
+    throw new Error(`Error: ${err}`)
```
bpierre, line 44: Should we add a bit more context here?

```suggestion
    throw new Error(`Could not describe transaction: ${err}`)
```

packages/connect-core/src/utils/metadata.ts
```diff
@@ -39,3 +48,29 @@ export async function resolveMetadata(
...
+    data.manifest
+  )
+}
+
+export async function resolveArtifact(
+  data: AppData | RoleData
+): Promise<AragonArtifact> {
+  if (hasAppInfo(data.appId, 'apm')) {
+    return getApmInternalAppInfo(data.appId)
+  } else if (hasAppInfo(data.appId, 'aragon')) {
```
bpierre, line 67: Now that we are returning from a function, we could get rid of the `else if` here, and the `else` below.

yarn.lock
```diff
@@ -226,6 +226,16 @@
   dependencies:
```
bpierre, line 226: Wondering if there was a reason to update the yarn.lock in this PR?

0xGabi: No, I'll remove that.

packages/connect-core/src/utils/metadata.ts
0xGabi: Nice, doing that change.

PR #124, opened 2020-07-08
Add example to create an organization

0xGabi:


examples/nodejs/src/helpers.ts
```diff
@@ -1,7 +1,52 @@
...
-}
\ No newline at end of file
+}
+
+export async function fetchRepo(name: string, subgraph: string) {
+  // Construct the query to fetch
+  const QUERY = gql`
+  query {
+    repos(where:{
+      name: "${name}"
```
bpierre, line 18: Should we use a GraphQL query argument for this?

examples/nodejs/src/helpers.ts
```diff
@@ -1,7 +1,52 @@
...
+  selectedFilter: string,
+  templateContract: ethers.Contract,
+  transactionHash: string
+): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const filter = templateContract.filters[selectedFilter]()
+
+    templateContract.on(filter, (contractAddress, event) => {
+      if (event.transactionHash === transactionHash) {
+        resolve(contractAddress)
```
bpierre, line 48: Should we call `templateContract.off()` once we get the address?

PR #129, opened 2020-07-09
Fast builds

bpierre:
_builds on https://github.com/aragon/connect/pull/128_

Normal (~16s)             |  Fast (<1s)
-|-
 | 

Fast build are experimental, and doing the following:

- Use esbuild.
- Skip type checking.
- Skip generating declarations.
- Skip reinstalling dependencies.
- Skip removing the built files.
- Execute `esbuild` directly rather than through yarn / oao.
- Build everything twice: CJS + ESM (in parallel so it has little impact on the build time).

Another script has been added to only check & generate the types: `build:types`.


packages/connect-core/package.json
```diff
@@ -2,9 +2,14 @@
...
   "version": "0.4.1",
   "license": "LGPL-3.0-or-later",
-  "main": "dist/index.js",
+  "main": "dist-cjs/index.js",
+  "browser": "dist/index.js",
+  "modules": "dist/index.js",
   "types": "dist/index.d.ts",
+  "sideEffects": false,
   "scripts": {
+    "build:fast": "esbuild --platform=browser --sourcemap --loader:.ts=ts --format=esm --target=es2019 --outdir=./dist --bundle ./src/index.ts",
```
0xGabi, line 11: Wonder if an `esbuild.config` exist?

scripts/rebuild-packages.sh
```diff
@@ -14,13 +26,72 @@ if [ "$MODE" == "libs-only" ]; then
...
+  package=$1
+
+  echo "  ⏳@aragon/$package"
+
+  cd "packages/$package"
+
+  esbuild_cmd cjs & esbuild_cmd esm
+  wait
+
+  echo -e "\e[1A\e[K\u001b[32m  ✔️\u001b[0m @aragon/$package"
```
0xGabi, line 64: This line prints a bit weird to me even though I have unicode characters.

bpierre: Could you post a screenshot? :thinking:

Also do you usually have emojis in your terminal?

packages/connect-core/package.json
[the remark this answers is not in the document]
bpierre: No I don’t think there is anything like that, [but the params should be a little bit more readable now](https://github.com/aragon/connect/pull/129/files#diff-014517145c178a5dce7cbc8797191e74R40-R50) :smiley:

scripts/rebuild-packages.sh
0xGabi: My bad I miss this message. Here it is:

Screen Shot 2020-08-14 at 6 47 13 PM

PR #130, opened 2020-07-09
Handle stranger cases in TokenManager

ajsantander:
This PR **updates the tokens subgraph from v0.1.0 to v0.1.3** (which is stored in the schema, within the AragonInfo object).
This version of the subgraph is being deployed in https://thegraph.com/explorer/subgraph/aragon/aragon-tokens-mainnet right now, and a positive result needs to be tested before this is merged. Otherwise, this PR should not be merged and the subgraph rolled back to v0.1.0.

The stranger case is with the following entities:
  * Org: 0x5aAd137d8f7D2Dc6e1B2548c059B1483360bCC6a
  * App: 0x983dd5e5ce0f33774f49666e00d6a820ff8d8a07
  * Token: 0x0077Cd5Eca22e86177cc465633227652ff02F32d

Such entities produce a situation where the TokenManager app is not initialized by the time that the subgraph picks it up, and thus does not know its token. This is probably because the DAO was created from the CLI (not a template), and the token deployed and set manually. What's difficult about this situation is that the token will never enter the subgraph via the usual data sources, and thus never be linked to the TokenManager app!

An easy fix would be to add this TokenManager's token as a data source, but this would mean a complete re-index of the entire universal subgraph each time an "orphan" token like this is found. Alternatively, cases like this could add the token data source, remove a few other data sources, and deploy their own subgraph.

However, the way this PR attempts to fix the problem is by adding a new mechanism, where uninitialized TokenManagers are stored in a cache, and **whenever a new app is detected in the subgraph**, the list of pending-to-be-linked TokenManagers is checked. As soon as one of these can see its token, both are linked and the app is removed from the cache.

We tried triggering the orphan checking upon entering any new block, but unfortunately this makes the subgraph extremely slow. It's a nice feature, but not really production-friendly.


packages/connect-thegraph-tokens/subgraph/manifest/data/mainnet-staging.json
```diff
@@ -3,16 +3,22 @@
   "dataSources": {
     "Organizations": [
       {
-        "name": "PieDAO",
-        "address": "0x0c188B183FF758500D1D18B432313d10e9F6b8a4",
-        "startBlock": "9593674"
+        "name": "DeepDAO",
```
ajsantander, line 6: Here's where we set the parameters to test the stranger case in fast staging mode.

packages/connect-thegraph-tokens/subgraph/manifest/data/mainnet-staging.json
```diff
@@ -3,16 +3,22 @@
...
+        "address": "0x5aAd137d8f7D2Dc6e1B2548c059B1483360bCC6a",
+        "startBlock": "8134920"
       }
     ],
     "OrganizationFactories": [],
     "OrganizationTemplates": [],
     "LegacyOrganizationTemplates": [],
     "TokenFactories": [],
-    "Tokens": []
+    "Tokens": [
```
ajsantander, line 15: The MiniMeToken entity needs to be set as a separate data source and hope the app and token entity are gracefully linked asynchronously in subgraph's indexing.

packages/connect-thegraph-tokens/subgraph/schema.graphql
```diff
@@ -2,7 +2,7 @@ type TokenManager @entity {
   id: ID!
   address: Bytes!
   orgAddress: Bytes!
-  token: MiniMeToken! @derivedFrom(field: "tokenManager")
+  token: MiniMeToken @derivedFrom(field: "tokenManager")
```
ajsantander, line 5: A TokenManager's token can now be null, since it accounts for cases of a TokenManager app that has not been initialized.

packages/connect-thegraph-tokens/subgraph/src/TokenManager.ts
```diff
@@ -24,7 +24,9 @@ export function getTokenManagerEntity(proxyAddress: Address): TokenManagerEntity
     tokenManagerEntity.orgAddress = tokenManagerContract.kernel()
 
     let tokenAddress = tokenManagerContract.token()
-    aragon.processToken(tokenAddress)
+    if (tokenAddress.toHexString() != '0x0000000000000000000000000000000000000000') {
```
ajsantander, line 27: Here's where we don't set the app's token address if it is not set yet.

packages/connect-thegraph-tokens/subgraph/src/aragon/aragon.ts
```diff
@@ -25,10 +25,12 @@ export function processApp(appAddress: Address, appId: string): void {
   }
 }
 
-export function processToken(tokenAddress: Address): void {
+export function processToken(tokenAddress: Address, createTemplate: boolean = true): void {
```
ajsantander, line 28: The AragonInfo registry can now register a token whose template has already been created.

packages/connect-thegraph-tokens/subgraph/src/aragon/mappings/MiniMeToken.ts
```diff
@@ -50,6 +49,8 @@ function _getMiniMeTokenEntity(previousBlock: BigInt, tokenAddress: Address): Mi
 
   let miniMeTokenEntity = MiniMeTokenEntity.load(miniMeTokenEntityId)
   if (!miniMeTokenEntity) {
+    aragon.processToken(tokenAddress, false)
```
ajsantander, line 52: If the incoming token is an orphan, make sure that its registered in the AragonInfo registry.

packages/connect-thegraph-tokens/subgraph/src/aragon/aragon.ts
```diff
@@ -80,7 +82,7 @@ function _getAragonInfo(): AragonInfoEntity {
   if (!aragon) {
     aragon = new AragonInfoEntity(aragonId)
 
-    aragon.version = 'tokens-v0.1.0'
+    aragon.version = 'tokens-v0.1.1'
```
ajsantander, line 85: This is where we're tracking subgraph versions for now.

packages/connect-thegraph-tokens/subgraph/src/aragon/mappings/MiniMeToken.ts
sohkai: Might be useful to leave a comment here to explain this!

packages/connect-thegraph-tokens/subgraph/manifest/templates/contracts/Kernel.template.yaml
```diff
@@ -39,4 +39,6 @@
     eventHandlers:
       - event: NewAppProxy(address,bool,bytes32)
         handler: handleNewAppProxy
+    blockHandlers:
```
ajsantander, line 42: Kernel data sources now trigger actions on new blocks. Yay!

This should work on all networks, and we need to monitor if this makes indexing slower. I don't think it should.

packages/connect-thegraph-tokens/subgraph/schema.graphql
```diff
@@ -25,6 +25,11 @@ type TokenHolder @entity {
   balance: BigInt!
 }
 
+type OrphanTokenManagers @entity {
```
ajsantander, line 28: This is where we cache TokenManagers that are not linked to a token yet.

packages/connect-thegraph-tokens/subgraph/src/TokenManager.ts
```diff
@@ -23,15 +24,81 @@ export function getTokenManagerEntity(proxyAddress: Address): TokenManagerEntity
...
+      _registerOrphanTokenManager(proxyAddress)
+    }
 
     tokenManagerEntity.save()
   }
 
   return tokenManagerEntity!
 }
 
+export function processOrphanTokenManagers(): void {
```
ajsantander, line 42: This chunk of code is where the mechanism lives :D

packages/connect-thegraph-tokens/subgraph/src/aragon/aragon.ts
```diff
@@ -1,9 +1,13 @@
-import { Address, DataSourceTemplate } from '@graphprotocol/graph-ts'
+import { ethereum, Address, DataSourceTemplate } from '@graphprotocol/graph-ts'
 import { AragonInfo as AragonInfoEntity } from '../../generated/schema'
 import { Kernel as KernelTemplate } from '../../generated/templates'
 import { MiniMeToken as MiniMeTokenTemplate } from '../../generated/templates'
 import * as hooks from '../aragon-hooks'
 
+export function processBlock(block: ethereum.Block): void {
```
ajsantander, line 7: New hook entry point!

packages/connect-thegraph-tokens/subgraph/src/aragon/mappings/MiniMeToken.ts
```diff
@@ -111,3 +102,22 @@ function _getTokenHolder(previousBlock: BigInt, miniMeTokenEntity: MiniMeTokenEn
 
   return tokenHolder
 }
+
+function _getTotalSupplyAt(tokenContract: MiniMeTokenContract, blockNumber: BigInt): BigInt {
```
ajsantander, line 106: Thought it'd be nicer to process calls that could fail in separate helper functions.

packages/connect-thegraph-tokens/subgraph/src/TokenManager.ts
```diff
@@ -23,15 +24,81 @@ export function getTokenManagerEntity(proxyAddress: Address): TokenManagerEntity
...
+  }
+}
+
+function _unregisterOrphanTokenManager(appAddress: Address): void {
+  let registry = _getOrphanTokenManagersEntity()
+
+  let apps = registry.apps
+  if (apps.includes(appAddress)) {
+    let idx = apps.indexOf(appAddress)
+    apps.splice(idx, 1)
```
0xGabi, line 79: I wonder if we can do this kind of logic on AssemblyScript?

packages/connect-thegraph-tokens/subgraph/src/aragon-hooks.ts
```diff
@@ -25,3 +28,7 @@ export function onAppTemplateCreated(appAddress: Address, appId: string): void {
 export function onOrgTemplateCreated(orgAddress: Address): void {}
 
 export function onTokenTemplateCreated(tokenAddress: Address): void {}
+
+export function onBlockEntered(block: ethereum.Block): void {
```
0xGabi, line 32: I like how you solve this adding a new hook!

packages/connect-thegraph-tokens/subgraph/manifest/templates/contracts/Kernel.template.yaml
sohkai: Trying to understand this mechanism a bit better—we are essentially asking to be notified on each new block that we see?

We place this handler on the kernel because we have no where better to put it?

ajsantander: Looks like it is making things slower. Needs optimization.

packages/connect-thegraph-tokens/subgraph/src/TokenManager.ts
ajsantander: Looks like it works, yeah. This is a trace I did to verify (newest in the bottom):
```
> Unregister tm, Orphans(before): [[object Object]], data_source: DeepDAO, block_hash: 0x5fb5fe5f291530cfffc81267e849fde388462cb436b1d71987f120ac8cdaa12a, block_number: 8134972
> Unregister tm, Orphans(after): [], data_source: DeepDAO, block_hash: 0x5fb5fe5f291530cfffc81267e849fde388462cb436b1d71987f120ac8cdaa12a, block_number: 8134972
> Register tm, Orphans(after): [[object Object]], data_source: DeepDAO, block_hash: 0xbd23eb97c7de5ffc998f19148748ecca4249ddc0bab35916e1b699a5706728d2, block_number: 8134920
> Register tm, Orphans(before): [], data_source: DeepDAO, block_hash: 0xbd23eb97c7de5ffc998f19148748ecca4249ddc0bab35916e1b699a5706728d2, block_number: 8134920
```

packages/connect-thegraph-tokens/subgraph/manifest/templates/contracts/Kernel.template.yaml
ajsantander: After discussing with The Graph, there doesn't seem to be a way to accelerate block handling. Even after reducing the amount of triggers that cause a block to be processed, its still very slow. So, unfortunately, we'll have to forego this option.

PR #131, opened 2020-07-11
Example: Create Gardens Template 

0xGabi:


0xGabi:
We can remove this example for now as we already have another example of creating a Honey/Karma organization using Connect.  Once the issue we were experiencing with the deployment is figuring out we can consider adding it as a new example.

Issue #109, opened 2020-07-03
Rename “readProvider” to “ethereum”


bpierre:
Or “provider”.

PR #135, opened 2020-07-14
connectApp()

bpierre:
_Fixes https://github.com/aragon/connect/issues/109_
_Builds on https://github.com/aragon/connect/pull/136, https://github.com/aragon/connect/pull/137_

## App connection

This PR adds a new way to connect apps:

```tsx
import { connect } from '@aragon/connect'
import connectVoting from '@aragon/connect-voting'

const org = await connect('myorg.aragonid.eth', 'thegraph')
const voting = await org.app('voting')
const votingConnected = await connectVoting(voting)
const votes = await votingConnected.votes()
```

`connectVoting()` accepts either an `App` or a promise resolving to an `App`, allowing this syntax:

```tsx
import { connect } from '@aragon/connect'
import connectVoting from '@aragon/connect-voting'

const myorg  = await connect('myorg.aragonid.eth', 'thegraph')
const voting = await connectVoting(myorg.app('voting'))
const votes  = await voting.votes()
```

## Usage from React

This is how the same task can get accomplished using the React library:

```jsx
import { createAppHook, useApp } from '@aragon/connect-react'
import connectVoting from '@aragon/connect-voting'

// We need to create a hook corresponding to the app connector
const useVoting = createAppHook(connectVoting)

function Votes() {
  const [voting] = useApp('voting')
  const [votes] = useVoting(voting, app => app.votes())
  return (
    <ul>
      {votes ? (
        votes.map(vote => <li key={vote.id}>{formatVote(vote)}</li>)
      ) : (
        <li>Loading votes…</li>
      )}
    </ul>
  )
}
```

Note: this doesn’t support subscriptions yet. I plan to introduce a separate change in the app connectors that will allow to subscribe using the exact same syntax as above.

## App connector authoring

This is how an app connector can get implemented in the most basic way:

```tsx
import { createAppConnector } from '@aragon/connect-core'

export default createAppConnector(() => ({
  // This object is going to get extended by the original App object.
  total: () => fetchTotal()
}))
```

And this is how the Voting app connector is implemented:

```tsx
import { createAppConnector } from '@aragon/connect-core'
import Voting from './entities/Voting'
import VotingConnectorTheGraph, {
  subgraphUrlFromChainId,
} from './thegraph/connector'

type Config = {
  subgraphUrl: string
}

// The app and Config types are optional but if passed,
// they will be properly forwarded to the returned function.

export default createAppConnector<Voting, Config>(
  ({ app, config, connector, network, verbose }) => {
    if (connector !== 'thegraph') {
      console.warn(
        `Connector unsupported: ${connector}. Using "thegraph" instead.`
      )
    }

    return new MyAppConnector(
      new VotingConnectorTheGraph(
        config.subgraphUrl ?? subgraphUrlFromChainId(network.chainId),
        verbose
      ),
      app.address
    )
  }
)
```

## Status

- [x] Make the app connectors universal
  - [x] Refactor Voting to use the new API + support multiple connectors.
  - [x] Refactor Tokens to use the new API + support multiple connectors.
  - [x] Rename `@aragon/connect-thegraph-voting` to `@aragon/connect-voting`.
  - [x] Rename `@aragon/connect-thegraph-tokens` to `@aragon/connect-tokens`.
- [x] Update documentation.
- [x] Find a way to tell TS about the connected app type.
- [x] Provide a utility function that lets authors build an app connector by extending `App`.
- [x] `@aragon/connect-react` compatibility.

## Other changes

- Export `connect` as a default export of `@aragon/connect`.
- Add a `toNetwork()` function that transforms `Networkish` into `Network`.
- `Organization` now receives a single `ConnectionContext` object.
- Prepare the ipfs resolver from the options, injects it into `ConnectionContext`.
- `provider` => `ethersProvider`, `readProvider` => `ethereumProvider`.
- Move the initialization steps into `connect()` rather than doing it in `Organization`.
- `Organization`: remove the `_connect()` method.
- Add new demos: `list-votes-cli` and `list-votes-react`.
- Update the documentation of app connectors (and remove subgraph / internal documentation from it).


packages/connect/src/connect.ts
```diff
@@ -156,15 +163,9 @@ async function connect(
...
   const orgConnector = getConnector(connector, network)
 
-  // Two independent tasks in parallel are done here:
-  //  - Resolving the organization address.
-  //  - Calling .connect() on the connector.
-  const [orgAddress] = await Promise.all([
-    resolveAddress(ethersProvider, location),
-    orgConnector.connect?.(),
-  ])
+  const orgAddress = await resolveAddress(ethersProvider, location)
```
0xGabi, line 166: One thing we are still missing is during the `getEthersProvider` we should manually create a WebSocket provider for Rinkeby because of the issue we saw that the default provider is not overriding our custom `ensAddress` correctly.

0xGabi: Note that the example with rinkeby is not working for this reason.

0xGabi: Open this issue on ethers repo: https://github.com/ethers-io/ethers.js/issues/959

bpierre: It’s working now, thanks for reporting the issue :pray:  https://github.com/aragon/connect/pull/135/commits/78c0c21acb831cc24556d0b73278a17c78b34768

packages/connect-core/src/entities/App.ts
```diff
@@ -2,14 +2,16 @@ import Repo from './Repo'
...
   AragonArtifact,
   AragonManifest,
+  ConnectionContext,
   Metadata,
-  Abi,
-  AppIntent,
 } from '../types'
 import { resolveManifest, resolveArtifact } from '../utils/metadata'
 import IOrganizationConnector from '../connections/IOrganizationConnector'
+// import IAppConnected from '../connections/IAppConnected'
```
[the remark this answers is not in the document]
bpierre: I ended up removing it, and let app connectors extend `App` themselves :+1:

packages/connect-core/src/connections/IAppConnected.ts
```diff
@@ -0,0 +1 @@
+export default interface IAppConnected {}
```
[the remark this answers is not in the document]
bpierre: I removed both of them since app connectors are now extending `App`, but the idea was that `IAppConnected` would have been the wrapping object (e.g. [`Voting`](https://github.com/aragon/connect/blob/d7518138e48e8200801670db8f30ec671ea71e3d/packages/connect-thegraph-voting/src/entities/Voting.ts)), while `IAppConnector` would have been the internal connector class (e.g. [`VotingConnectorTheGraph`](https://github.com/aragon/connect/blob/d7518138e48e8200801670db8f30ec671ea71e3d/packages/connect-thegraph-voting/src/connector.ts)).

examples/list-votes-cli/src/index.ts
```diff
@@ -0,0 +1,61 @@
+import { connect } from '@aragon/connect'
+import votingConnector from '@aragon/connect-thegraph-voting'
+
+const BLUE = '\x1b[36m'
+const RESET = '\x1b[0m'
+
+const env = {
```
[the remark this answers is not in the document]
bpierre: Done in https://github.com/aragon/connect/pull/135/commits/a81c665db73e3a76488cee85832a6563a5b32aa1

packages/connect-voting/src/__test__/casts.test.ts
```diff
@@ -1,25 +1,25 @@
 import { VotingConnectorTheGraph, Vote, Cast } from '../../src'
 
-const VOTING_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/aragon/aragon-voting-rinkeby'
+const VOTING_SUBGRAPH_URL =
```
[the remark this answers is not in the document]
bpierre: Yes I think we could definitely improve the way we test things now that objects are instantiated by the library itself. Maybe the JSON provider could be useful here too. Do you see things we could do now, or should we talk about it separately?

packages/connect/src/connect.ts
```diff
@@ -1,34 +1,60 @@
...
   ConnectorTheGraphConfig,
 } from '@aragon/connect-thegraph'
-import { Network } from '@aragon/connect-types'
+import { Address, Network, Networkish } from '@aragon/connect-types'
 
-export type ConnectOptionsResolveIpfs = (
-  ipfsIdentifier: string,
-  path: string
-) => string
+const XDAI_WSS_ENDPOINT = 'wss://xdai.poanetwork.dev/wss'
```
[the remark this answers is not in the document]
bpierre: `params.js` is in `@aragon/connect-core` while we are in `@aragon/connect` here, I was finding it a bit weird to have a constant in a library not using it. I created a `constants.js` in `@aragon/connect`!

docs/connectors/voting-app.md
```diff
@@ -1,92 +1,42 @@
...
 
-```javascript
-import { Voting } from '@aragon/connect-thegraph-voting'
+```js
+import connect from '@aragon/connect'
+import connectVoting from '@aragon/connect-voting'
 
-const voting = new Voting(VOTING_APP_ADDRESS, VOTING_APP_SUBGRAPH_URL)
+const org = connect('myorg.aragonid.eth', 'thegraph')
+const voting = connectVoting(org.app('voting'))
```
Evalir, line 14: It's so short now! 🏎️

examples/connect-react-intro/tsconfig.json
```diff
@@ -1,11 +1,14 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "@snowpack/app-scripts-react/tsconfig.base.json",
```
Evalir, line 2: Snowpack 👁️ ❄️

examples/nodejs/src/create-karma-template.ts
```diff
@@ -0,0 +1,133 @@
...
+  DECAY,
+  MAX_RATIO,
+  BigNumber.from(WEIGHT.toString()),
+]
+
+async function main() {
+  // create signer
+  const provider = new ethers.providers.JsonRpcProvider(env.rpc, env.network)
+  const wallet = new ethers.Wallet(env.key)
+  const signer = wallet.connect(provider)
```
Evalir, line 66: Should we print the public key of the wallet?

examples/nodejs/src/helpers.ts
```diff
@@ -29,9 +29,11 @@ export async function fetchRepo(name: string, subgraph: string) {
   const wrapper = new GraphQLWrapper(subgraph)
 
   // Invoke the custom query and receive data
-  return wrapper.performQuery(QUERY_REPO_BY_NAME, {
+  const { data } = await wrapper.performQuery(QUERY_REPO_BY_NAME, {
     name,
   })
+
+  return data.repos[0]
```
Evalir, line 36: Hmmm, if this returns null (due to a network error or something of the sort), this may just explode. This is probably the reason that makes connectors break if we update our subgraphs as well. Should we prefix it with `?`, or should we just let the user handle it?

packages/connect-core/src/utils/misc.ts
```diff
@@ -1,3 +1,75 @@
...
+          `Please use one of the following: ${NETWORKS.map(
+            network => network.chainId
+          ).join(', ')}.`
+      )
+    }
+    return network
+  }
+
+  if (!value) {
+    throw new Error(`Network: incorrect value provided.`)
```
Evalir, line 52: Hmm, shouldn't we say something like `You must provide a "value" argument to the network config`?

packages/connect-core/src/utils/misc.ts
```diff
@@ -1,3 +1,75 @@
...
+  }
+
+  if (value.chainId === undefined) {
+    throw new Error(`Network: no chainId provided.`)
+  }
+
+  if (value.name === undefined) {
```
Evalir, line 59: This was probably for clarity :P But shouldn't we do `!value.chainId` / `!value.name`?

packages/connect-core/src/utils/misc.ts
bpierre: Yes we should have something like this at an upper level, but in `toNetwork()` there is no concept of configuration yet. I plan to add custom errors for the next version, which will make it much easier to do things like this! :+1:

packages/connect-core/src/utils/misc.ts
bpierre: This is to accept a `chainId` being `0` or a `name` being `""`, just in case someone does that :smile:

examples/nodejs/src/helpers.ts
bpierre: (this was part of another PR sorry :s)

It seems like it yes. We should definitely improve the reliability of the parts where data is fetched, it’s something I want to do soon alongside providing better errors. I also think we should consider `GraphQLWrapper` to be an object used internally or by app connectors, but not by end users or demos.

packages/connect-voting/src/__test__/casts.test.ts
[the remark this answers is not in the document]
0xGabi: Yes, let's first release version 0.5 and refactor the Intent API and afterwards we can concentrate fully on tests.

PR #136, opened 2020-07-14
Types: add SubscriptionHandler, Address

bpierre:
Extracting some changes out of https://github.com/aragon/connect/pull/135 to make it easier to review.


docs/api-reference/organization.md
```diff
@@ -25,7 +25,7 @@ Like `Organization#app()`, but as a subscription.
 | `appFilter.address` | `String`                      | Same as `appFilter`, but makes the selection by `address` explicit.                                                                                                                                                                                           |
```
0xGabi, line 25: I wonder what you think of using the types (`Address` and `SubscriptionHandler`) on the documentation as well. And maybe we should have a new documentation page with all the types we currently have?

bpierre: Yes definitely, I think we could have a page for the types somewhere. :+1:

PR #137, opened 2020-07-14
Rename ConnectorInterface to IOrganizationConnector

bpierre:
It also now contains the `name` and `network` properties.

### DIFF
--- a/packages/connect-thegraph/src/connector.ts
+++ b/packages/connect-thegraph/src/connector.ts
@@ -1,6 +1,6 @@
 import {
   App,
-  ConnectorInterface,
+  IOrganizationConnector,
   Permission,
   Repo,
   Role,
@@ -17,8 +17,9 @@ import {
 } from './parsers'
 
 export type ConnectorTheGraphConfig = {
-  verbose?: boolean
+  network: Network
   orgSubgraphUrl?: string
+  verbose?: boolean
 }
 
 function getOrgSubgraphUrl(network: Network): string | null {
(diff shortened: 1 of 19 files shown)


packages/connect-thegraph/src/connector.ts
```diff
@@ -50,31 +51,35 @@ function appFiltersToQueryFilter(appFilters: AppFilters) {
...
+
+    if (!orgSubgraphUrl) {
       throw new Error(
-        `The chainId ${network.chainId} is not supported by the TheGraph connector.`
+        `The chainId ${config.network.chainId} is not supported by the TheGraph connector.`
       )
     }
-    super(_orgSubgraphUrl, verbose)
+
+    this.#gql = new GraphQLWrapper(orgSubgraphUrl, config.verbose)
```
0xGabi, line 69: Nice! 🙌

bpierre:
> My only concern is with the backward compatibility issue we will introduce moving `GraphQLWrapper` to `#gql`. I think is the right decision but we should clearly state that the next version has breaking changes.

I was wondering if you saw code using the inherited `GraphQLWrapper` methods? I was considering it to be an internal API.

In terms of breaking changes, I was more thinking about the [constructor params](https://github.com/aragon/connect/commit/6456de95589d907bbad704c5a4f82d1affc8efc6#diff-7ce8ce67e000cc00c9c430036d6cd544R15-R17) of the connectors. But I don’t see a reason to instantiate `ConnectorTheGraph` manually since it’s embedded, so chances are that people won’t have to make any change here.

In any case, I 100% agree that we should document any breaking change (or any other change :smile:) in the public API :+1:

0xGabi:
No totally, you are right. I got confused by the inheritance. Developers are using the `GraphQLWrapper` directly and not from the `OrganizationConnector` so should no be backward compatibility issues other than what you mention about constructor params.

PR #140, opened 2020-07-14
Connectors: add a disconnect() method (fix hanging tests)

bpierre:
This closes the WS connections that were preventing the tests to run properly.

Note that connectors won’t have to declare a `connect()` and `disconnect()` methods as they are both optional.


packages/connect-thegraph/src/types.ts
```diff
@@ -1,6 +1,14 @@
...
 
 export type ParseFunction = (data: DataGql, connector: any) => {}
+
+// From https://github.com/FormidableLabs/urql/blob/ca68584a578b9f85d0b1448fc7a5fc9587f968de/packages/core/src/exchanges/subscription.ts#L39-L44
+export interface SubscriptionOperation {
+  query: string
+  variables?: object
+  key: string
+  context: OperationContext
+}
```
bpierre, line 14: This is to avoid importing a file using an internal path (`@urql/core/dist/types/exchanges/subscription`), and prevent it to break while also making it possible to follow different import paths based CJS / ESM support.

packages/connect-thegraph/src/connector.ts
```diff
@@ -65,6 +65,12 @@ export default class ConnectorTheGraph extends GraphQLWrapper
     super(_orgSubgraphUrl, verbose)
   }
 
+  async connect() {}
+
+  async disconnect() {
+    this.close()
```
0xGabi, line 71: Super clean solution!

packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -6,33 +6,38 @@ import {
...
 import { SubscriptionClient } from 'subscriptions-transport-ws'
 import { DocumentNode } from 'graphql'
-import { ParseFunction, QueryResult } from '../types'
 import { pipe, subscribe } from 'wonka'
+import { ParseFunction, QueryResult, SubscriptionOperation } from '../types'
+
+const AUTO_RECONNECT = true
+const CONNECTION_TIMEOUT = 20 * 1000
+
+function filterSubgraphUrl(url: string): [string, string] {
```
0xGabi, line 17: Nice!

packages/connect-thegraph/src/types.ts
0xGabi: Good choice.

packages/connect-thegraph/src/connector.ts
```diff
@@ -65,6 +65,12 @@ export default class ConnectorTheGraph extends GraphQLWrapper
     super(_orgSubgraphUrl, verbose)
   }
 
+  async connect() {}
```
0xGabi, line 68: I wonder what kind of feature you envision we can include in this method for the organization connector.

bpierre: On #135, it is used to store the connection, as a second step: https://github.com/aragon/connect/pull/135/files?file-filters%5B%5D=.ts#diff-7b005797d9aa8e707f98f0dc25a4017fR75

Doing it in two steps is necessary because the connection object contains the connector :smile:

PR #165, opened 2020-07-27
Agreement: Implement subgraph and initial entities

facuspagnuolo:
Testing deployed subgraph: [aragon-agreement-rinkeby-staging](https://thegraph.com/explorer/subgraph/facuspagnuolo/aragon-agreement-rinkeby-staging)


packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
```diff
@@ -0,0 +1,297 @@
...
+
+function buildDisputeId(agreement: Address, disputeId: BigInt): string {
+  return agreement.toHexString() + "-dispute-" + disputeId.toString()
+}
+
+function buildVersionId(agreement: Address, versionId: BigInt): string {
+  return agreement.toHexString() + "-version-" + versionId.toString()
+}
+
+function buildCollateralRequirementId(agreement: Address, disputable: Address, collateralRequirementId: BigInt): string {
```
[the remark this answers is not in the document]
facuspagnuolo: Good catch!

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
```diff
@@ -0,0 +1,297 @@
+import { ERC20 } from '../generated/schema'
+import { ERC20 as ERC20Contract } from '../generated/templates/Agreement/ERC20'
+import { ethereum, BigInt, Address } from '@graphprotocol/graph-ts'
+import { Agreement, Action, Signature, Version, Disputable, Challenge, Dispute, Evidence, Signer, CollateralRequirement } from '../generated/schema'
```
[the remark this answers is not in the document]
facuspagnuolo: I like the initiative! The only thing that could be confusing is the `AgreementEntity`, especially since it is not the same as the entities exposed by the connector, these are subgraph entities

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
```diff
@@ -0,0 +1,297 @@
...
+  action.createdAt = event.block.timestamp
+  action.save()
+}
+
+export function handleActionClosed(event: ActionClosed): void {
+  let agreementApp = AgreementContract.bind(event.address)
+  let actionData = agreementApp.getAction(event.params.actionId)
+
+  let action = Action.load(buildActionId(event.address, event.params.actionId))!
+  action.closed = actionData.value5
```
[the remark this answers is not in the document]
facuspagnuolo: I like the idea!

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
```diff
@@ -0,0 +1,297 @@
...
+
+  let challenge = new Challenge(challengeId)
+  let challengeData = agreementApp.getChallenge(event.params.challengeId)
+  challenge.action = buildActionId(event.address, event.params.actionId)
+  challenge.challengeId = event.params.challengeId
+  challenge.challenger = challengeData.value1
+  challenge.endDate = challengeData.value2
+  challenge.context = challengeData.value3
+  challenge.settlementOffer = challengeData.value4
+  challenge.arbitratorFeeAmount = challengeData.value5
```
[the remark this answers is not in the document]
facuspagnuolo: Yup, we merged a few changes last week, updating the subgraph now. I don't expect any breaking changes for the contract interface for now.

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
```diff
@@ -0,0 +1,297 @@
...
+}
+
+export function handleActionDisputed(event: ActionDisputed): void {
+  updateChallengeState(event.address, event.params.challengeId)
+
+  let agreementApp = AgreementContract.bind(event.address)
+  let challengeData = agreementApp.getChallenge(event.params.challengeId)
+
+  let dispute = new Dispute(buildDisputeId(event.address, challengeData.value10))
+  dispute.ruling = challengeData.value11
```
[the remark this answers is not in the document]
facuspagnuolo: Answered above!

packages/connect-thegraph-agreement/subgraph/scripts/deploy.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Arguments
+USER=$1
+NAME=$2
+NETWORK=$3
+
```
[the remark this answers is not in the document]
facuspagnuolo: True! I'd be nice to have all these scripts in an internal shared sub-repo, WDYT?

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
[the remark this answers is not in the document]
facuspagnuolo: Nice catch!

facuspagnuolo:
Thanks @0xGabi for the very thorough review! I addressed all your comments, let me know your thoughts!

packages/connect-agreement/subgraph/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
sohkai, line 1: Not directly related here, but I saw this in a few other subgraphs too and I'm wondering if we should remove it? It seems weird this would be MIT and copyright with The Graph.

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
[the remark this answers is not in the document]
0xGabi: Yes, that's a good point. In my opinion, having the distinction on the mappings clarify regardless. Maybe we can think on a better naming to the connectors entities?

packages/connect-agreement/src/entities/Signer.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.id = data.id
+    this.address = data.address
+    this.agreementId = data.agreementId
+  }
+
+  async disconnect() {
+    await this.#connector.disconnect()
+  }
+
+  async hasSigned(versionNumber: string): Promise<boolean> {
```
0xGabi, line 25: Leaving a comment here to remind ourselves of a good use-case of further data transformation that we can do at the Connector layer.

packages/connect-agreement/src/types.ts
```diff
@@ -19,13 +21,30 @@ export interface VersionData {
   effectiveFrom: string
 }
 
+export interface SignerData {
```
0xGabi, line 24: We should move the rest of the app connector to keep interfaces on the `types` file. 👍

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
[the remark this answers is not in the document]
facuspagnuolo: Yea I agree, what about `models` for the connector entities instead? It's just a synonym, but it will help to clarify

packages/connect-agreement/src/entities/Signature.ts
```diff
@@ -0,0 +1,19 @@
+import { SignatureData, IAgreementConnector } from '../types'
+
+export default class Signature {
+  #connector: IAgreementConnector
+
+  readonly id!: string
+  readonly signerId!: string
+  readonly versionId!: string
+  readonly createdAt!: string
```
bpierre, line 9: I think we could remove the `!`s here (they disable the non-null check, which shouldn’t be needed since we assign them in the constructor).

packages/connect-agreement/jest.config.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = {
+  ...baseConfig,
+  roots: [`<rootDir>/packages/${packageName}`],
+  collectCoverageFrom: ['src/**/*.{ts,tsx}'],
+  testRegex: `(packages/${packageName}/.*/__tests__/.*|\\.(test|spec))\\.tsx?$`,
+  modulePaths: [`<rootDir>/packages/${packageName}/src/`],
+  name: packageName,
+  displayName: 'AGREEMENT',
+  testTimeout: 30000,
```
bpierre, line 13: I’m wondering if this is still needed, now that we have a disconnection mechanism?

packages/connect-agreement/subgraph/LICENSE
0xGabi: I agree I'll create a new PR once we merge this one to remove it everywhere.

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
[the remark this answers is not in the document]
0xGabi: Great, I like it. I can refactor to the new naming on a new PR.

packages/connect-agreement/src/entities/Agreement.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  async disconnect() {
+    await this.#connector.disconnect()
+  }
+
+  async id(): Promise<string> {
+    const data = await this.#connector.agreement(this.#appAddress)
+    return data.id
+  }
```
bpierre, line 23: Just a note on not having `onX` equivalents for some of the methods: I think it makes sense, some methods don’t need to receive updates so it would be a bit odd to ask connector authors to implement them.

It would still be useful to always be able to subscribe (so we could add e.g. a retry mechanism), and to provide a unified API, but this is something we can do at a higher level, e.g. in `connect-core`.

packages/connect-agreement/src/entities/Signer.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import Signature from '../entities/Signature'
+import { SignerData, IAgreementConnector } from '../types'
+
+export default class Signer {
+  #connector: IAgreementConnector
+
+  readonly id!: string
+  readonly address!: string
+  readonly agreementId!: string
```
bpierre, line 11: We could also remove the `!`s here:
```suggestion
  readonly id: string
  readonly address: string
  readonly agreementId: string
```

packages/connect-agreement/src/entities/Signer.ts
```diff
@@ -0,0 +1,38 @@
...
+    const versionId = `${this.agreementId}-version-${versionNumber}`
+    const signatures = await this.signatures()
+    return signatures.some((signature: Signature) => signature.versionId === versionId)
+  }
+
+  async signatures({ first = 1000, skip = 0 } = {}): Promise<Signature[]> {
+    return this.#connector.signatures(this.id, first, skip)
+  }
+
+  onSignatures(callback: Function): SubscriptionHandler {
```
bpierre, line 35: We should have the `signatures()` filters as a first parameter here.

packages/connect-agreement/src/entities/Version.ts
```diff
@@ -0,0 +1,25 @@
...
+export default class Version {
+  #connector: IAgreementConnector
+
+  readonly id!: string
+  readonly versionId!: string
+  readonly content!: string
+  readonly title!: string
+  readonly arbitrator!: string
+  readonly appFeesCashier!: string
+  readonly effectiveFrom!: string
```
bpierre, line 12: We should remove the `!` :gun: here as well.

packages/connect-agreement/src/thegraph/connector.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  async versions(agreement: string, first: number, skip: number): Promise<Version[]> {
+    return this.#gql.performQueryWithParser(
+      queries.ALL_VERSIONS('query'),
+      { agreement, first, skip },
+      (result: QueryResult) => parseVersions(result, this)
+    )
+  }
+
+  onVersions(agreement: string, callback: Function): SubscriptionHandler {
```
bpierre, line 90: The `first: number` and `skip: number` parameters from `versions()` are missing here. Should we also make them a filter object?

packages/connect-agreement/src/thegraph/connector.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  async signatures(signerId: string, first: number, skip: number): Promise<Signature[]> {
+    return this.#gql.performQueryWithParser(
+      queries.GET_SIGNATURES('query'),
+      { signerId, first, skip },
+      (result: QueryResult) => parseSignatures(result, this)
+    )
+  }
+
+  onSignatures(signerId: string, callback: Function): SubscriptionHandler {
```
bpierre, line 124: Same than `onVersions()`: we should align the parameters with `signatures()`, and probably move `first` and `skip` in a filters object.

packages/connect-agreement/src/thegraph/parsers/agreement.ts
```diff
@@ -0,0 +1,18 @@
+import { QueryResult } from '@aragon/connect-thegraph'
+
+import { AgreementData } from '../../types'
+
+export function parseAgreement(result: QueryResult): AgreementData {
+  const agreement = result.data.agreement
+
+  if (!agreement) {
```
bpierre, line 8: Should we start having validating functions for the data we receive from The Graph (e.g. doing `isValid(agreement)` rather than `!agreement`) in the `parsers/*` files? For example we rely on `agreement.currentVersion` to be an object below, but it might not be the case.

packages/connect-agreement/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": "src"
+  },
+  "include": ["./src/**/*.ts"],
```
bpierre, line 7:
```suggestion
  "include": ["src"],
```

packages/connect-agreement/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "dist",
```
bpierre, line 4:
```suggestion
    "outDir": "dist/esm",
```

yarn.lock
```diff
@@ -2,6 +2,13 @@
 # yarn lockfile v1
 
 
+"@aragon/connect-react@^0.4.2":
+  version "0.4.2"
+  resolved "https://registry.yarnpkg.com/@aragon/connect-react/-/connect-react-0.4.2.tgz#8f6a4cc88c8f52d5fe762dd21d92fe37e8e8e685"
+  integrity sha512-oQ67hLTfDhzLJRDvmGRoabDMCi9Nvi+o29MQ/swvDroPHs8EHtW5PAPndZjga69pS3Th12D8sGcFp+CNDePV/Q==
+  dependencies:
+    "@aragon/connect" "*"
+
```
bpierre, line 11: Should we remove this change from the PR?

packages/connect-thegraph-agreement/subgraph/src/Agreement.ts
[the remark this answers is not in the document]
bpierre: Just wanted to confirm that we are only referring to subgraph-related code here? I would avoid using suffixes for entities [at the library level](https://github.com/aragon/connect/tree/agreement_subgraph/packages/connect-agreement/src/entities) since these are exposed to users.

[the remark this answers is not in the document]
facuspagnuolo: Yup, my intention was to use synonyms for the subgraph entities, but also rename the `entities` dir in the connector to `models` to avoid confusing these

packages/connect-agreement/src/thegraph/parsers/agreement.ts
facuspagnuolo: Good point, I agree we can definitely do better here! Let's discuss in general what's the ideal pattern that could work for these types of validations. I'll be happy to address it in a separate PR :)

facuspagnuolo:
Thanks for the review @bpierre! I addressed almost all your comments and answered others, let me know your thoughts!

packages/connect-agreement/src/thegraph/parsers/agreement.ts
bpierre: Agreed, it would make sense to do this in all the app connectors at once :+1:

bpierre:
> Thanks for the review @bpierre! I addressed almost all your comments and answered others, let me know your thoughts!

All good for me! :pray: :100:

PR #178, opened 2020-08-03
DisputableVoting: Implement subgraph

facuspagnuolo:
You can try the subgraph here: [aragon-dvoting-rinkeby-staging](https://thegraph.com/explorer/subgraph/facuspagnuolo/aragon-dvoting-rinkeby-staging) (there's a limit for the name's length)


packages/connect-voting-disputable/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@aragon/connect-voting-disputable",
```
bpierre, line 2: Should we use the name `@aragon/connect-disputable-voting`? It seems to be a more common way to refer to it.

packages/connect-voting-disputable/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "dist",
```
bpierre, line 4: Just to align with the other packages ([it is also specified in the package.json](https://github.com/aragon/connect/pull/178/files#diff-70f05e0405597eaccd6b66bce74da065R11)):

```suggestion
    "outDir": "dist/esm",
```

packages/connect-voting-disputable/tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+    "rootDir": "src"
+  },
+  "include": ["./src/**/*.ts"],
```
bpierre, line 7: It seems that specifying the directory only is enough (I moved the other packages to do the same):

```suggestion
  "include": ["src"],
```

packages/connect-voting-disputable/subgraph/src/DisputableVoting.ts
```diff
@@ -0,0 +1,213 @@
...
+  vote.setting = buildSettingId(event.address, voteData.value3)
+  vote.actionId = voteData.value4
+  vote.status = castVoteStatus(voteData.value5)
+  vote.startDate = voteData.value6
+  vote.snapshotBlock = voteData.value7
+  vote.pausedAt = voteData.value8
+  vote.pauseDuration = voteData.value9
+  vote.quietEndingExtendedSeconds = voteData.value10
+  vote.quietEndingSnapshotSupport = castVoterState(voteData.value11)
+  vote.script = voteData.value12
```
0xGabi, line 70: I wonder if the latest version of the contract is here: https://github.com/aragon/aragon-apps/blob/disputable_voting_deploy/apps/voting-disputable/contracts/DisputableVoting.sol#L85

I notice the `voteData` maybe have small differences.

facuspagnuolo: Yup, I double-checked and it is that one :)
We are also checking it in the tests

PR #194, opened 2020-08-07
DisputableVoting: Extend vote model functionalities

facuspagnuolo:
This PR adds the following functionalities for the `Vote` model:
- Expose vote duration
- Compute end date
- Vote results
- Querying the collateral requirements
- Knowing whether a voter has voted or not


packages/connect-voting-disputable/src/models/Vote.ts
```diff
@@ -48,11 +52,50 @@ export default class Vote {
...
+
+  get yeasPct(): string {
+    return this.votingPowerPct(this.yeas)
+  }
+
+  get naysPct(): string {
+    return this.votingPowerPct(this.nays)
+  }
+
+  votingPowerPct(num: string): string {
```
0xGabi, line 69: Nice!

packages/connect-voting-disputable/src/models/Vote.ts
```diff
@@ -48,11 +52,50 @@ export default class Vote {
     this.script = data.script
   }
 
+  get endDate(): string {
```
0xGabi, line 55: Leaving a comment to point a good utility on the Connectors layer.

packages/connect-voting-disputable/src/models/CollateralRequirement.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  readonly id: string
+  readonly voteId: string
+  readonly token: string
+  readonly actionAmount: string
+  readonly challengeAmount: string
+  readonly challengeDuration: string
+
+  constructor(data: CollateralRequirementData, connector: IDisputableVotingConnector) {
+    this.#connector = connector
```
0xGabi, line 14: I wonder about having the `connector` object on every model. Maybe we should only include it on those models that implement queries/subscriptions requests?

facuspagnuolo: Yes, I wasn't sure about that, but it makes sense, we can have it only if needed

packages/connect-voting-disputable/src/models/Vote.ts
facuspagnuolo: It would be nice to have a better way to handle bignumber ops to be honest, a few helpers would be wonderful

0xGabi: Wonder what you have in mind? Maybe make sense to have a `connect-utils`, @bpierre what you think?

PR #196, opened 2020-08-12
DisputableVoting: Expose vote settings and results

facuspagnuolo:
This PR adds the following concepts to the disputable voting app:
- Allow accessing ERC20's related info: name, symbol, and decimals
- Knowing whether a vote is passing or not
- New status labels for votes including "Rejected" and "Accepted"
- New vote attributes: challenger, challenge end date, challenge ID, dispute ID, and execution date
- Allow querying the setting associated with a vote
- Formatted values for all vote results: yeas, nays, yeasPct, naysPct
- Formatted values for all pct settings: support and acceptance

It would be nice to think about how we want to start structuring the logic related to helpers/utils, specially around times, numbers, etc. Maybe we can do sth with our existing [contract-helpers](https://github.com/aragon/contract-helpers/) (cc @bpierre @0xGabi)


packages/connect-voting-disputable/src/helpers/numbers.ts
```diff
@@ -0,0 +1,27 @@
+import { BigNumber, utils } from 'ethers'
+
+export const PCT_DECIMALS = 16 // 100% = 10^18
+
+export const PCT_BASE = BigNumber.from(`100${'0'.repeat(PCT_DECIMALS)}`)
+
+export const bn = (x: string): BigNumber => BigNumber.from(x)
+
+export const formatBn = (number: string | BigNumber, numberDecimals: string | number, formattedDecimals = 2): string => {
```
0xGabi, line 9: Wonder if we may wan to have `PCT_DECIMALS` as default value for `numberDecimals`.

packages/connect-voting-disputable/src/helpers/numbers.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (decimalPosition === - 1) {
+    return `${formattedNumber}.${'0'.repeat(formattedDecimals)}`
+  }
+
+  const decimals = formattedNumber.substring(decimalPosition + 1)
+  const decimalsLength = decimals.length
+
+  if (decimalsLength <= formattedDecimals) {
+    return `${formattedNumber}${'0'.repeat(formattedDecimals - decimalsLength)}`
+  }
```
0xGabi, line 22: I'm not super sure about this logic. I think we are missing to round the decimals when `decimalsLength <= formattedDecimals`.

packages/connect-voting-disputable/src/models/Setting.ts
```diff
@@ -28,4 +26,12 @@ export default class Setting {
     this.quietEndingExtension = data.quietEndingExtension
     this.createdAt = data.createdAt
   }
+
+  get formattedSupportRequiredPct(): string {
+    return formatBn(this.supportRequiredPct, PCT_DECIMALS)
```
0xGabi, line 31: If we decided to use `PCT_DECIMALS` as default we can remove it in a few places. Or do you think is better to be explicit about it?

packages/connect-voting-disputable/src/models/Vote.ts
```diff
@@ -50,12 +62,18 @@ export default class Vote {
     this.quietEndingExtendedSeconds = data.quietEndingExtendedSeconds
     this.quietEndingSnapshotSupport = data.quietEndingSnapshotSupport
     this.script = data.script
+    this.executedAt = data.executedAt
+  }
+
+  get hasEnded(): boolean {
+    const now = new Date().getTime()
+    return now >= toMilliseconds(this.endDate)
```
0xGabi, line 70: Nice!

packages/connect-voting-disputable/src/models/Vote.ts
```diff
@@ -50,12 +62,18 @@ export default class Vote {
...
   }
 
   get endDate(): string {
-    const originalEndDate = BigNumber.from(this.startDate).add(BigNumber.from(this.duration))
-    const endDateAfterPause = originalEndDate.add(BigNumber.from(this.pauseDuration))
-    return endDateAfterPause.add(BigNumber.from(this.quietEndingExtendedSeconds)).toString()
+    const originalEndDate = bn(this.startDate).add(bn(this.duration))
+    const endDateAfterPause = originalEndDate.add(bn(this.pauseDuration))
+    return endDateAfterPause.add(bn(this.quietEndingExtendedSeconds)).toString()
```
0xGabi, line 76: Wonder if we also want to convert the values with `toMilliseconds` so the consumer should always expect milliseconds from the connectors.

packages/connect-voting-disputable/src/models/Vote.ts
```diff
@@ -67,8 +85,45 @@ export default class Vote {
...
 
   votingPowerPct(num: string): string {
-    const votingPower = BigNumber.from(this.votingPower)
-    return BigNumber.from(num).mul(BigNumber.from(100)).div(votingPower).toString()
+    const votingPower = bn(this.votingPower)
+    return bn(num).mul(PCT_BASE).div(votingPower).toString()
+  }
+
+  async formattedYeas(): Promise<string> {
+    return formatBn(this.yeas, await this._tokenDecimals())
```
0xGabi, line 93: What you think of having the `tokenDecimals` be part of the Vote entity at the subgraph level. I wonder if we may prevent a lot of request to fetch the decimals value that's something we are sure is not going to change after the creation of the token.

packages/connect-voting-disputable/src/models/Vote.ts
```diff
@@ -67,8 +85,45 @@ export default class Vote {
...
+  }
+
+  async formattedYeasPct(): Promise<string> {
```
0xGabi, line 96:
```suggestion
formattedYeasPct(): string {

File: packages/connect-voting-disputable/src/models/Vote.ts [hunk 8]
```diff
@@ -67,8 +85,45 @@ export default class Vote {
...
+    return formatBn(this.yeasPct, PCT_DECIMALS)
+  }
+
+  async formattedNays(): Promise<string> {
+    return formatBn(this.nays, await this._tokenDecimals())
+  }
+
+  async formattedNaysPct(): Promise<string> {
```
0xGabi, line 104:
```suggestion
formattedNaysPct():string {

File: packages/connect-voting-disputable/src/models/Vote.ts [hunk 9]
```diff
@@ -98,4 +153,25 @@ export default class Vote {
...
+  _disputableVoting(): DisputableVoting {
+    return new DisputableVoting(this.#connector, this.votingId)
+  }
+
+  async _tokenDecimals(): Promise<string> {
+    const erc20 = await this._disputableVoting().token()
+    return erc20.decimals
+  }
+
+  _isValuePct(value: BigNumber, total: BigNumber, pct: BigNumber): boolean {
```
0xGabi, line 174: What you think if change the naming of this function? I got a bit confuse in relation to what it does.

packages/connect-voting-disputable/src/models/Vote.ts
facuspagnuolo: I like the idea :)

packages/connect-voting-disputable/src/models/Setting.ts
facuspagnuolo: Well, actually note that we are using different decimals for these operations, 16 for percentages and 18 for token amounts, then I'd keep it

packages/connect-voting-disputable/src/helpers/numbers.ts
facuspagnuolo: Note that scenario is contemplated in the tests, and actually, if the number of decimals is lower than or equal to the number of requested decimals to format, then we don't need to round it, don't you agree?

packages/connect-voting-disputable/src/helpers/numbers.ts
facuspagnuolo: Answered bellow in https://github.com/aragon/connect/pull/196#discussion_r472235428

packages/connect-voting-disputable/src/helpers/index.ts
```diff
@@ -0,0 +1,2 @@
+export { toMilliseconds } from './time'
```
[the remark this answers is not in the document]
facuspagnuolo: Yup, I already mentioned this to @sohkai... ideally I think we can start exporting some of the helpers with typescript support. However, we could probably start doing that internally for these connect packages and then export it to the contract-helpers repo right?

packages/connect-voting-disputable/src/models/Vote.ts
facuspagnuolo: Yes, I think that's a great idea. Although we could change all the models expose in connect at once to make sure there is no hybrid behavior, what do you think?

facuspagnuolo:
@0xGabi thx for the review! I addressed some of your comments and answered the rest of them. 
OTOH, I'm having an issue after merging the latest changes from master, maybe you are aware of what that could be about

packages/connect-voting-disputable/src/models/Vote.ts
```diff
@@ -50,25 +63,55 @@ export default class Vote {
...
     this.quietEndingSnapshotSupport = data.quietEndingSnapshotSupport
     this.script = data.script
+    this.executedAt = data.executedAt
+    this.isAccepted = data.isAccepted
+    this.tokenDecimals = data.tokenDecimals
+  }
+
+  get hasEnded(): boolean {
+    const now = new Date().getTime()
+    return now >= toMilliseconds(this.endDate)
```
sohkai, line 73: Note that you can also use `Date.now()` :)

packages/connect-voting-disputable/src/helpers/numbers.ts
```diff
@@ -0,0 +1,27 @@
+import { BigNumber, utils } from 'ethers'
```
sohkai, line 1: I wonder if it would be worth exposing `TokenAmount`s from app connectors' API (cc @bpierre).

(Also see next comment)

packages/connect-voting-disputable/src/helpers/index.ts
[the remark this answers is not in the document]
sohkai: I'd be against re-using the helpers we have in `contract-helpers-test`, to keep that package singularly focused on making it easier for us to write tests.

We can definitely think about a `connect-helpers` package for some of these, especially as we standardize on what is a good, well-formed app connector API—I would love for us to be in a place where app connectors are implementing their dates, numbers, and token amounts in a singular, well-recommended fashion that's obvious for developers to use.

packages/connect-voting-disputable/src/__test__/herlpers/numbers.test.ts
```diff
@@ -0,0 +1,49 @@
+import { bn, formatBn } from '../../helpers'
```
sohkai, line 1: Nitpick: the directory (`herlpers`—both kind of funny but scary when I initially read it as `herpies`) is typoed.

facuspagnuolo:
@0xGabi added some more stuff! let me know your thoughts!

packages/connect-voting-disputable/src/models/Vote.ts
0xGabi: Good point, let's do that!

packages/connect-voting-disputable/subgraph/src/DisputableVoting.ts
```diff
@@ -145,6 +146,16 @@ export function handleChangeRepresentative(event: ChangeRepresentativeEvent): vo
   voter.save()
 }
 
+function isAccepted(yeas: BigInt, nays: BigInt, votingPower: BigInt, settingId: string, pctBase: BigInt): boolean {
```
0xGabi, line 149: Look much better to manipulate the `BigNumber`s at the subgraph level 👍

PR #202, opened 2020-08-12
toNetwork(): make the name optional

bpierre:
Also moved the network-related functions into network.ts.


packages/connect-core/src/utils/network.ts
```diff
@@ -0,0 +1,83 @@
...
+          `Please use provide a chainId or use one of the following names: ` +
+          NETWORKS.map((network) => network.chainId).join(', ') +
+          `.`
+      )
+    }
+  }
+
+  // Just a little help for TypeScript, at this
+  // point we know that chainId cannot be undefined.
+  chainId = chainId as number
```
0xGabi, line 44: Nice!

PR #203, opened 2020-08-14
Replace GraphQL subscriptions by queries + polling

bpierre:
The Graph is having issues with GraphQL subscriptions, so we are moving to queries with polling for the time being. Once subscriptions get enabled again on The Graph, we will move back to GraphQL subscriptions.

Also in this PR:
- `nodejs/` demos update.
- Add the possibility to consume the organization connector from any app   connector, allowing to retrieve its options for example. This was done so that app connectors using TheGraph can inherit from pollInterval defined in the main connector.

### DIFF
--- a/packages/connect-agreement/src/thegraph/connector.ts
+++ b/packages/connect-agreement/src/thegraph/connector.ts
@@ -28,16 +28,25 @@ export function subgraphUrlFromChainId(chainId: number) {
   return null
 }
 
+type AgreementConnectorTheGraphConfig = {
+  pollInterval?: number
+  subgraphUrl?: string
+  verbose?: boolean
+}
+
 export default class AgreementConnectorTheGraph implements IAgreementConnector {
   #gql: GraphQLWrapper
 
-  constructor(subgraphUrl: string, verbose: boolean = false) {
-    if (!subgraphUrl) {
+  constructor(config: AgreementConnectorTheGraphConfig) {
+    if (!config.subgraphUrl) {
       throw new Error(
         'AgreementConnectorTheGraph requires subgraphUrl to be passed.'
       )
     }
-    this.#gql = new GraphQLWrapper(subgraphUrl, verbose)
+    this.#gql = new GraphQLWrapper(config.subgraphUrl, {
+      pollInterval: config.pollInterval,
+      verbose: config.verbose,
+    })
   }
 
   async disconnect() {
--- a/packages/connect-agreement/src/thegraph/queries/index.ts
+++ b/packages/connect-agreement/src/thegraph/queries/index.ts
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag'
 
 export const GET_AGREEMENT = (type: string) => gql`
-  ${type} Agreement($agreement: String!) {
+  query Agreement($agreement: String!) {
     agreement(id: $agreement) {
       id
       dao
@@ -14,7 +14,7 @@ export const GET_AGREEMENT = (type: string) => gql`
 `
 
 export const GET_CURRENT_VERSION = (type: string) => gql`  
-  ${type} Agreement($agreement: String!) {
+  query Agreement($agreement: String!) {
     agreement(id: $agreement) {
       currentVersion {
         id
@@ -30,7 +30,7 @@ export const GET_CURRENT_VERSION = (type: string) => gql`
 `
 
 export const GET_VERSION = (type: string) => gql`  
-  ${type} Version($versionId: String!) {
+  query Version($versionId: String!) {
     version(id: $versionId) {
       id
       versionId
@@ -44,7 +44,7 @@ export const GET_VERSION = (type: string) => gql`
 `
 
 export const ALL_VERSIONS = (type: string) => gql`
-  ${type} Versions($agreement: String!, $first: Int!, $skip: Int!) {
+  query Versions($agreement: String!, $first: Int!, $skip: Int!) {
     versions(where: {
       agreement: $agreement
     }, first: $first, skip: $skip) {
@@ -60,7 +60,7 @@ export const ALL_VERSIONS = (type: string) => gql`
 `
 
 export const GET_SIGNER = (type: string) => gql`
-  ${type} Signer($signerId: String!) {
+  query Signer($signerId: String!) {
     signer(id: $signerId) {
       id
       address
@@ -70,7 +70,7 @@ export const GET_SIGNER = (type: string) => gql`
 `
 
 export const GET_SIGNATURES = (type: string) => gql`
-  ${type} Signatures($signerId: String!, $first: Int!, $skip: Int!) {
+  query Signatures($signerId: String!, $first: Int!, $skip: Int!) {
     signatures(where: { 
       signer: $signerId
     }, first: $first, skip: $skip) {
--- a/packages/connect-agreement/src/thegraph/queries/index.ts
+++ b/packages/connect-agreement/src/thegraph/queries/index.ts
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag'
 
 export const GET_AGREEMENT = (type: string) => gql`
-  query Agreement($agreement: String!) {
+  ${type} Agreement($agreement: String!) {
     agreement(id: $agreement) {
       id
       dao
@@ -14,7 +14,7 @@ export const GET_AGREEMENT = (type: string) => gql`
 `
 
 export const GET_CURRENT_VERSION = (type: string) => gql`  
-  query Agreement($agreement: String!) {
+  ${type} Agreement($agreement: String!) {
     agreement(id: $agreement) {
       currentVersion {
         id
@@ -30,7 +30,7 @@ export const GET_CURRENT_VERSION = (type: string) => gql`
 `
 
 export const GET_VERSION = (type: string) => gql`  
-  query Version($versionId: String!) {
+  ${type} Version($versionId: String!) {
     version(id: $versionId) {
       id
       versionId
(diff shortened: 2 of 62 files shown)


packages/connect-agreement/src/thegraph/queries/index.ts
```diff
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag'
 
 export const GET_AGREEMENT = (type: string) => gql`
-  ${type} Agreement($agreement: String!) {
+  query Agreement($agreement: String!) {
```
Evalir, line 4: Hmm, why are we changing the `type` to query? I guess it renders the type param useless, right?

packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -61,7 +34,10 @@ export default class GraphQLWrapper {
     const request = createRequest(query, args)
 
     return pipe(
-      this.#client.executeSubscription(request),
+      this.#client.executeQuery(request, {
+        pollInterval: POLL_INTERVAL,
+        requestPolicy: 'cache-and-network',
```
bpierre, line 39: This [request policy](https://formidable.com/open-source/urql/docs/basics/queries/#request-policies) makes urql serve the cache first, and fetch the result again from the network. Network wise it is as efficient as having no cache, but for users, it will make cached queries update instantly, while updating the content right after if the query result has changed.

packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -61,7 +34,10 @@ export default class GraphQLWrapper {
     const request = createRequest(query, args)
 
     return pipe(
-      this.#client.executeSubscription(request),
+      this.#client.executeQuery(request, {
+        pollInterval: POLL_INTERVAL,
```
bpierre, line 38: I am wondering if we should have an option for this on `connect()`?

For example:

```
const org = await connect('myorg', 'thegraph', { pollInterval: 20000 })
```

The day we move back to GraphQL subscriptions, we could make it `-1` by default, so that any change on this value could keep using polling (it might be useful to support both).

packages/connect-agreement/src/thegraph/queries/index.ts
bpierre: > Hmm, why are we changing the type to query?

The GraphQL endpoint errors if we try to perform a query with `subscription` in the query itself.

>  I guess it renders the type param useless, right?

I was trying to keep things as similar as possible to avoid having to ask connectors to adapt… but we will have to do it anyway. Thanks for pointing this out, let me check again to see if we can change the query from the object generated by `gql` so that connectors using `GraphQLWrapper` wouldn’t have to change anything. :+1:

bpierre: Done in https://github.com/aragon/connect/pull/203/commits/5a44409e03580b7d82280783370feeff6d6fd91d

packages/connect-thegraph/src/core/GraphQLWrapper.ts
Evalir: This seems pretty sound; we definitely don't want stale state, but getting instant responses if there's a cache sounds great!

packages/connect-thegraph/src/core/GraphQLWrapper.ts
Evalir: Yeah, having this as a configurable option sounds good! I was thinking that we should also could make this standard for the connectors, so we can have granularity over which connectors we may want to set a polling interval.

packages/connect-thegraph/src/core/GraphQLWrapper.ts
```diff
@@ -1,67 +1,63 @@
...
-  return [url.replace(/^wss/, 'https'), url.replace(/^https/, 'wss')]
+  return createRequestUrql(query, args)
 }
 
 export default class GraphQLWrapper {
   #client: Client
   #verbose: boolean
-  close: () => void
 
   constructor(subgraphUrl: string, verbose = false) {
```
0xGabi, line 34: Maybe having a new argument to provide the `POLL_INTERVAL` that the GraphQLWrapper will use.

packages/connect-thegraph/src/core/GraphQLWrapper.ts
0xGabi: Yes, I think it makes sense to have it as an option on `connect()`. I also agree with Enrique suggestion of having it on connectors. As mentioned above maybe adding it as a new argument on the GraphQLWrapper constructor?

packages/connect-thegraph/src/core/GraphQLWrapper.ts
0xGabi: Good decisions, being able to update the urlq Client cache on every attempt request would be great for The Graph.

bpierre:
> On a separate note: I'm guessing there are other examples out of date, right?

Indeed! All the examples should be up to date now: https://github.com/aragon/connect/pull/205

packages/connect-thegraph/src/core/GraphQLWrapper.ts
bpierre: Done in https://github.com/aragon/connect/pull/203/commits/171161d904116902fbacc801d7eaeb7566d20888

packages/connect-voting-disputable/src/__test__/models/votes.test.ts
```diff
@@ -1,13 +1,21 @@
-import { DisputableVoting, Vote, CastVote, DisputableVotingConnectorTheGraph } from '../../../src'
+import {
```
0xGabi, line 1: I notice we had several lint corrections on a few of the files. What you think of adding workflow with a tool like husky to improve the developer practice and only allow push after linting?

packages/connect-agreement/src/thegraph/connector.ts
```diff
@@ -28,16 +28,25 @@ export function subgraphUrlFromChainId(chainId: number) {
   return null
 }
 
+type AgreementConnectorTheGraphConfig = {
```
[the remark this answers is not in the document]
bpierre: Do you mean for our connectors only, or as a type that would be exported by `connect-core` for any connector to extend the base config? The way I see it, a connector configuration is the responsibility of the connector defining it, even if they happen to be identical in our case because we follow the same conventions. What do you think?

packages/connect-core/src/connections/IOrganizationConnector.ts
```diff
@@ -9,6 +9,7 @@ import Role from '../entities/Role'
 export default interface IOrganizationConnector {
   readonly name: string
   readonly network: Network
+  readonly config: any
```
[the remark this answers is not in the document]
bpierre: I couldn’t use `object` or `{}` because TS will enforce that they stay empty… but I found another way :smile: e78a90ed826659c389477368461fb893dc0230d1

packages/connect-finance/src/connect.ts
```diff
@@ -1,27 +1,33 @@
-import { createAppConnector, Network } from '@aragon/connect-core'
-import { IFinanceConnector } from './types'
+import { createAppConnector } from '@aragon/connect-core'
 import Finance from './models/Finance'
 import FinanceConnectorTheGraph, {
   subgraphUrlFromChainId,
 } from './thegraph/connector'
 
 type Config = {
```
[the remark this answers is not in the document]
bpierre: I guess in our case we might want to have a base that we would extend, but I also like the idea of keeping them separate, even if it require some repetition, because it makes it easy for connector authors to take one and use it as a bootstrap. What do you think?

packages/connect-tokens/src/connect.ts
```diff
@@ -1,27 +1,37 @@
...
+      config.subgraphUrl ?? subgraphUrlFromChainId(network.chainId) ?? undefined
+
+    let pollInterval
+    if (orgConnector.name === 'thegraph') {
+      pollInterval =
+        config?.pollInterval ?? orgConnector.config?.pollInterval ?? undefined
+    }
+
+    const tokensConnector = await TokensConnectorTheGraph.create({
+      appAddress: app.address,
```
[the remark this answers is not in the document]
bpierre: Agreed, I made them all separate. 0d332ccca1b32cb693787bc90169c761752bc291

packages/connect-tokens/src/thegraph/connector.ts
```diff
@@ -29,21 +36,28 @@ export default class TokensConnectorTheGraph implements ITokensConnector {
...
+
+    const gql = new GraphQLWrapper(config.subgraphUrl, {
+      pollInterval: config.pollInterval,
+      verbose: config.verbose,
+    })
 
     const token = await gql.performQueryWithParser<Token>(
       queries.TOKEN('query'),
-      { tokenManagerAddress: appAddress },
+      { tokenManagerAddress: config.appAddress },
```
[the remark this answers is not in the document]
bpierre: Using `create()` is needed in the tokens app connector, because it’s the only one that need to instantiate asynchronously. What would you think of moving all the connectors to an async `create()`, even if they don’t need it (yet)?

packages/connect-voting-disputable/src/__test__/models/votes.test.ts
bpierre: Yes good idea! I am going to add this as a separate PR.

bpierre: https://github.com/aragon/connect/pull/207

PR #205, opened 2020-08-14
Examples cleanup

bpierre:
Three main things:

- Move from webpack 5 to webpack 4 (webpack 5 is having some issues with ethers at the moment).
- Add bundler-specific demos (Parcel and Webpack with ts-loader).
- Use latest Connect APIs.

### DIFF
--- a/examples/org-viewer-react/package.json
+++ b/examples/org-viewer-react/package.json
@@ -1,29 +1,26 @@
 {
   "name": "org-viewer-react",
-  "main": "dist/index.js",
   "version": "0.5.3",
   "private": true,
-  "sideEffects": false,
   "scripts": {
-    "dev": "yarn clean && webpack-dev-server",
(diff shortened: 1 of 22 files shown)


examples/org-viewer-react/package.json
```diff
@@ -1,29 +1,26 @@
 {
   "name": "org-viewer-react",
-  "main": "dist/index.js",
   "version": "0.5.3",
   "private": true,
-  "sideEffects": false,
```
0xGabi, line 6: Why we remove the `sideEffects` flag? Is it because of webpack v4?

examples/org-viewer-react/src/environment.ts
```diff
@@ -22,8 +21,6 @@ const ENVIRONMENTS = new Map([
     'mainnet_staging',
```
0xGabi, line 21: Should we remove the staging environment as we did for the `org-viewer-web` example?

PR #213, opened 2020-08-18
Docs: Remove features we don't support yet

### LINKED ISSUE #180: Doc: Review and remove feature we do not support yet
The documentation has a few features we have in our roadmap but that do not land to production yet. This is confusing for the current users of the library. For example the `options.as` on the `TransactionIntent` object.

0xGabi:
Close #180


bpierre (approved):
:+1:

Issue #132, opened 2020-07-11
API changes: Intents, Forwarding Paths and Transactions


bpierre:
This issue will keep track of the planned changes regarding the intents, forwarding paths and transactions.

These changes are aiming to improve / solve the following:

1. Separate the forwarding paths from the transactions to sign [1].
2. Provide a standard transaction request object that can be passed to Ethereum libraries.
3. Make it easier to sign a single transaction.
4. Make it possible to ignore the forwarding paths.
5. Make it possible to express complex forwarding paths.

[1] This is why the term “Forwarding Path” is now preferred to “Transaction Path”.

## Current Status

1. [ ] Store the signing address in the context (WIP, @bpierre)
2. [ ] Finalize the path builder API
3. [ ] Finalize the path declaration syntax
4. [ ] Reshape the intent-to-transaction flow
5. [ ] Implement the path builder (requires 2 and 4)
6. [ ] Implement the path declaration syntax (requires 3 and 4)

## Storing the signing address in the connection context

The idea here is to attach a default address to the connection context. This address will get used by the methods generating forwarding paths and transaction requests. It will be possible to override it at any point.

It could be passed to `connect()`:

```ts
const myorg = await connect('myorg.aragonid.eth', 'thegraph', {
  actAs: '0x…',
})
```

It should also be possible to update it without resetting the connection. A new method on `Organization` could help doing that:

```ts
org.actAs('0x…')
```

### `Intent`

- Replaces `AppIntent`.
- Represents an action to be executed.
- Only represents the intent, the forwarding path is not calculated at this point.
- Returned by some methods of `Organization` and `App`.
- Users should not instantiate it directly (using `new`).

```ts
class Intent {

  // Retrieve a single forwarding path. Defaults to the shortest one.
  path(PathOptions): Promise<ForwardingPath>

  // Retrieve the different possible forwarding paths.
  paths(PathOptions): Promise<ForwardingPath[]>

  // A list of transaction requests ready to get signed.
  transactions(PathOptions): Promise<Transaction[]>

  // Lets consumers pass a callback to sign any number of transactions.
  // This is similar to calling transactions() and using a loop, but shorter.
  // It returns the value returned by the library, usually a transaction receipt.
  sign<Receipt>(
    callback: (tx: Transaction) => Promise<Receipt>,
    options: PathOptions
  ): Promise<Receipt[]>
}
```

#### Example

```ts
const org = await connect('myorg', { account })
const voting = org.app('voting')

[...]

PR #214, opened 2020-08-19
Intent API refactor

0xGabi:
Address "Implement the intent-to-transaction API changes." from #132


0xGabi:
This PR was split in several and more defined list of changes: #223 #224 #225 #226

PR #223, opened 2020-08-20
API refactor: Types + Transaction + ForwardingPath

0xGabi:
Closes #212

PR #224, opened 2020-08-20
API Refactor: Transaction

### LINKED ISSUE #220: API Refactor: Transactions
This issue keeps track of the changes that need to happen to the `Transaction` entity described on #132.

- [x] Update the data types on every place
- [x] Refactor how we handle the creation of the Transaction entities during the Forwarding Path calculation

0xGabi:
Closes #220 

Builds on #223


packages/connect-core/src/entities/App.ts
```diff
@@ -59,6 +61,15 @@ export default class App {
     return this.organization.connection.orgConnector
   }
 
+  interface(): ethers.utils.Interface {
```
0xGabi, line 64: A helper to access an Ethers interface object from the app entities.

packages/connect-core/src/utils/transactions.ts
```diff
@@ -55,42 +16,17 @@ export async function createDirectTransactionForApp(
   methodSignature: string,
   params: any[]
 ): Promise<Transaction> {
-  if (!app) {
-    throw new Error(`Could not create transaction due to missing app artifact`)
-  }
+  const appInterface = app.interface()
```
0xGabi, line 19: We refactor the logic to use Ethers interface directly instead of having the old logic to search for the `methodAbiFragment` manually.

packages/connect-core/src/utils/transactions.ts
```diff
@@ -1,52 +1,13 @@
...
-
-export async function createDirectTransaction(
-  sender: string,
-  destination: string,
-  methodAbiFragment: ethers.utils.FunctionFragment,
-  params: any[]
-): Promise<Transaction> {
-  let transactionOptions = {}
-
-  // If an extra parameter has been provided, it is the transaction options if it is an object
```
0xGabi, line 29: We don't longer need to support the transaction options as overrides as we did on aragonAPI. Now, this is the responsibility of the users of Connect when they sign the transaction we provide.

packages/connect-core/src/utils/transactions.ts
```diff
@@ -181,16 +116,13 @@ export async function buildForwardingFeePretransaction(
...
   if (feeDetails.tokenAddress && feeDetails.amount > BigInt(0)) {
     // Needs a token approval pretransaction
-    const forwardingTxWithTokenData: TransactionWithTokenData = {
-      ...forwardingTransaction,
-      token: {
-        address: feeDetails.tokenAddress,
-        spender: forwarderAddress, // since it's a forwarding transaction, always show the real spender
-        value: feeDetails.amount.toString(),
-      },
+    const tokenData: TokenData = {
```
0xGabi, line 119: We no longer pass the Token details as part of the transaction object instead we provide a new object.

packages/connect-core/src/utils/descriptions.ts
```diff
@@ -1,76 +1,76 @@
-import { ethers } from 'ethers'
```
0xGabi, line 1: To prevent errors with types I decided to comment the logic related to the Forwarding Path descriptions.

PR #225, opened 2020-08-20
Refactor: Add Intent entity + refactor calculate path logic

### LINKED ISSUE #217: API refactor: Intent logic
This issue keeps track of the changes that need to happen related to the `Intent` entity described on #132.

- [ ] New methods to interact with `Transaction`: `sign`, `transactions`
- [ ] New method to calculate all the possible Forwarding Paths: `paths`
- [ ] Update the current `path` method to use the new data structures and types

0xGabi:
Closes #217 

Builds on #224


packages/connect-core/src/entities/App.ts
```diff
@@ -90,11 +90,11 @@ export default class App {
     return this.artifact.abi
   }
 
-  get intents(): AppIntent[] {
+  get methods(): AppMethod[] {
```
0xGabi, line 93: I decide to change the name to `methods` instead of `intents` as it was confusing. I'm also wondering if we may want to remove those methods at all 🤔

packages/connect-core/src/entities/App.ts
```diff
@@ -61,6 +61,19 @@ export default class App {
     return this.organization.connection.orgConnector
   }
 
+  contract(): ethers.Contract {
```
0xGabi, line 64: I include a helper to have access to an Ethers contract object to make it super easy to interact with the blockchain for the particular App.

packages/connect-core/src/entities/Intent.ts
```diff
@@ -0,0 +1,81 @@
...
+  ) {
+    this.#org = org
+    this.#provider = provider
+    this.appAddress = data.appAddress
+    this.functionName = data.functionName
+    this.functionArgs = data.functionArgs
+  }
+
+  // Retrieve a single forwarding path. Defaults to the shortest one.
+  async path({ actAs, path }: PathOptions): Promise<TransactionPath> {
```
0xGabi, line 29: We removed all the logic related to the description of a path and handle that responsibility to the Forwarding Path.

packages/connect-core/src/entities/Intent.ts
```diff
@@ -0,0 +1,81 @@
...
+          .some((address) => address === app.address)
+      ),
+      destination,
+      transactions,
+    })
+  }
+
+  // Retrieve the different possible forwarding paths.
+  async paths({ actAs, path }: PathOptions): Promise<TransactionPath[]> {
+    // TODO: support logic to calculate multiple Forwarding paths
```
0xGabi, line 62: We left the calculation of multiple Forwarding Path for a later PR.

packages/connect-core/src/utils/path/calculatePath.ts
```diff
@@ -53,37 +23,53 @@ async function calculateForwardingPath(
...
-    return { path: [] }
+    return []
   }
 
   const createForwarderTransaction = createForwarderTransactionBuilder(
     sender,
     directTransaction
   )
 
+  const buildForwardingPath = async (
```
0xGabi, line 37: Extracted this logic as it was used multiple times.

packages/connect-core/src/utils/path/calculatePath.ts
```diff
@@ -181,34 +150,26 @@ async function calculateForwardingPath(
  */
 export async function calculateTransactionPath(
   sender: string,
-  destination: string,
+  destinationApp: App,
```
0xGabi, line 153: We now accept an App entity instead.

packages/connect-core/src/utils/path/calculatePath.ts
```diff
@@ -181,34 +150,26 @@ async function calculateForwardingPath(
...
-    method.sig,
+    methodSignature,
     params
   )
 
+  const finalForwarderProvided = finalForwarder
+    ? ethers.utils.isAddress(finalForwarder)
+    : false
+
+  const method = getAppMethod(destinationApp, methodSignature)
```
0xGabi, line 172: We are only using the `method` to know if it's protected by a role. I wonder if we can get this information in other way 🤔

packages/connect-core/src/utils/path/calculatePath.ts
```diff
@@ -249,11 +210,13 @@ export async function calculateTransactionPath(
       (role) => role.name === method.roles[0]
     )
     const allowedEntities =
-      role?.permissions?.map((permission) => permission.granteeAddress) || []
+      role?.permissions
+        ?.filter((permission) => permission.allowed)
```
0xGabi, line 214: We missed to check if the permission was allowed.

packages/connect-core/src/utils/path/calculatePath.ts
```diff
@@ -53,37 +23,53 @@ async function calculateForwardingPath(
   forwardersWithPermission: string[],
   forwarders: string[],
   provider: ethers.providers.Provider
-): Promise<PathData> {
+): Promise<Transaction[]> {
   // No forwarders can perform the requested action
   if (forwardersWithPermission.length === 0) {
-    return { path: [] }
+    return []
```
0xGabi, line 29: We return an array the Transaction object.

Issue #219, opened 2020-08-20
API refactor: Describe a Forwarding Path


0xGabi:
This issue goes further into the details of what it takes to create a Forwarding Path description. The logic that we currently support was migrated from `aragon.js` and it's not flexible enough to provide lot of value for custom UI experiences.

To transition into a more powerful and descriptive data type as the one described in #132 we need to address a few tasks:

- [ ] Update the core logic that handles how the `ForwardingPathDescription` is constructed
- [ ] Refactor how we use Radspec to provide more flexibility and control to Connect users
- [ ] Consider moving the helpers and utilities to encode and decode descriptions to their own Connect package

PR #226, opened 2020-08-20
Refactor: Forwarding Path

### LINKED ISSUE #218: API refactor: Forwarding Path
This issue keep track of all the tasks that need to happen to refactor the Forwarding Path described on #132.

- [x] New methods to interact with `Transaction`: `sign`, `transactions`
- [ ] New method to describe a Forwarding Path. The logic to construct a description will change a lot. So this item needs further refinement that we will address on a separate issue:

0xGabi:
Closes #218 

Builds on #225 

Note: This is still a draft PR, needs to work better on the logic of how to describe a `ForwardingPath`.


packages/connect-core/src/entities/ForwardingPath.ts
```diff
@@ -0,0 +1,77 @@
...
+  // It returns the value returned by the library, usually a transaction receipt.
+  async sign<Receipt>(
+    callback: (tx: Transaction) => Promise<Receipt>
+  ): Promise<Receipt[]> {
+    return Promise.all(this.transactions.map(async (tx) => await callback(tx)))
+  }
+
+  // Return a description of the forwarding path, to be rendered.
+  async describe(): Promise<ForwardingPathDescription> {
+    // TODO: Make sure we are safe to only provide the apps on the path here
```
0xGabi, line 37: Here I decide to use the apps that form the path but I wonder if we might need to know the whole list of installed apps. If that case we need to construct the `ForwardingPath` with a reference to the organization so we can fetch information like the list of apps.

packages/connect-core/src/entities/ForwardingPath.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export type ForwardingPathDescriptionTree = ForwardingPathDescriptionTreeEntry[]
+
+export class ForwardingPathDescription {
+  readonly apps: App[]
+  readonly describeSteps: PostProcessDescription[]
+
+  constructor(data: ForwardingPathDescriptionData) {
+    this.apps = data.apps
+    this.describeSteps = data.describeSteps
```
0xGabi, line 60: I'm still not super sure about the data structure we would like to have for the `ForwardingPathDescription`. I just decided to use a list of `PostProcessDescription` objects. Those are returned by Radspec. By we may probably want to further think about it.

packages/connect-core/src/entities/Organization.ts
```diff
@@ -134,4 +136,10 @@ export default class Organization {
       this.connection.ethersProvider
     )
   }
+
+  //////// DESCRIPTIONS /////////
+
+  async describeScript(script: string): Promise<ForwardingPathDescription> {
```
0xGabi, line 142: I decided to expose the `describeScript` utility to the organization as we already have all the contextual information we need and user and app connectors can leverage directly.

packages/connect-core/src/utils/description.ts
```diff
@@ -0,0 +1,94 @@
+import { ethers } from 'ethers'
```
0xGabi, line 1: Here is the main logic we are currently using to both encode and decode an evm script and transform it into a `ForwardingPathDescrption`. I'm not convinced yet about the current data types and how we constructing the description but I think we are going in the right direction. Maybe it's a big topic and we should address it as a separate PR and further discuss it in: https://github.com/aragon/connect/issues/219

PR #229, opened 2020-08-23
Move to Node-style callbacks in every subscription

bpierre:
This is a breaking change for app connectors and consumers (with the exception of `@aragon/connect-react`). It is necessary to pass errors with subscriptions, the same way we do it for the async functions by throwing errors.

A compatible solution could be to pass the error as a second parameter, but many people are used to Node-style callbacks, so I think it would be too weird to invert the parameters.

Contains the following changes:

- Add the `SubscriptionCallback` type.
- Add types to all the subscriptions methods.
- Move all the subscription methods to use node-style callbacks.
- Update documentation and examples.

### DIFF
--- a/packages/connect-thegraph/src/connector.ts
+++ b/packages/connect-thegraph/src/connector.ts
@@ -168,10 +168,7 @@ class ConnectorTheGraph implements IOrganizationConnector {
         orgAddress: organization.address.toLowerCase(),
       },
       callback,
-      async (result) => {
-        const apps = await parseApps(result, organization)
-        return apps[0]
-      }
+      (result) => parseApp(result, organization)
     )
   }
 
--- a/packages/connect-thegraph/src/connector.ts
+++ b/packages/connect-thegraph/src/connector.ts
@@ -14,6 +14,7 @@ import {
   Network,
   Networkish,
   SubscriptionHandler,
+  SubscriptionCallback,
 } from '@aragon/connect-types'
 import * as queries from './queries'
 import GraphQLWrapper from './core/GraphQLWrapper'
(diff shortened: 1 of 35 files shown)


packages/connect-thegraph/src/connector.ts
```diff
@@ -154,41 +155,44 @@ class ConnectorTheGraph implements IOrganizationConnector {
...
         appFilter: appFiltersToQueryFilter(filters),
+        first: 1,
         orgAddress: organization.address.toLowerCase(),
       },
-      (result) => parseApps(result, organization)
+      callback,
+      async (result) => {
+        const apps = await parseApps(result, organization)
+        return apps[0]
+      }
```
0xGabi, line 174:
```suggestion
      (result) => parseApp(result, organization)

File: packages/connect-thegraph/src/core/GraphQLWrapper.ts [hunk 2]
```diff
@@ -76,31 +79,52 @@ export default class GraphQLWrapper {
...
-              this.describeQueryResult(result),
-            ].join('\n')
+          callback(
+            new Error(
+              'Error performing subscription.\n' +
+                `${result.error.name}: ${result.error.message} \n` +
+                this.describeQueryResult(result)
+            )
           )
+          return
```
0xGabi, line 90: Would you mind explain why do we need to return here?

0xGabi:
Merge this one to make #233 easier to review.

packages/connect-thegraph/src/connector.ts
```diff
@@ -168,10 +168,7 @@ class ConnectorTheGraph implements IOrganizationConnector {
         orgAddress: organization.address.toLowerCase(),
       },
       callback,
-      async (result) => {
-        const apps = await parseApps(result, organization)
-        return apps[0]
-      }
+      (result) => parseApp(result, organization)
```
bpierre, line 171: @0xGabi Wondering if you tested this? Because the query is `ORGANIZATION_APPS` with a limit of `1` since we added the filters (the single app query was only working by address).

packages/connect-thegraph/src/core/GraphQLWrapper.ts
bpierre: It is so that it stops here in case of error, and doesn’t continue to call `callback(null, result)` which is right after.

packages/connect-thegraph/src/connector.ts
0xGabi: Apology, I didn't check on my local the change as it was small and seems straight forward but I got confused with how we were handling `APP_BY_ADDRESS` query.

Issue #216, opened 2020-08-20
React: Votes returning null inbetween poll updates


andy-hook:
This might be related to the changes in https://github.com/aragon/connect/pull/203

In between polling updates, `votes` return `null` before being repopulated.

Here is an example from the react demo:

![image](https://user-images.githubusercontent.com/11708259/90747036-93d88a80-e2c8-11ea-894d-467655f0bb35.png)

This causes jumps between loading and display at every interval.

Is this expected behaviour?

PR #233, opened 2020-08-24
React API improvements

bpierre:
_fixes https://github.com/aragon/connect/issues/216, builds on https://github.com/aragon/connect/pull/229_

This PR contains various improvements for the React API. I suggest to review them commit by commit, every section ends with a link to the related commits (“see …”).

## Loading behavior

Since we replaced the GraphQL subscriptions by queries with polling (see #203), the hooks provided by the React API were updating too often, and incorrectly (see #216). This version is now taking care of only triggering updates when a changed happened, and to only use loading states when necessary.

See 0f0206a93bbb4caf84abfec538a0dc77a5faabe7, a223382da8f31d9b3843dcc419c3fe8096d247f2, 35898096ed0cdd2e8c76ac77ea65b02a49a2f81b.

## createAppHook(): allow connector configuration to be set

App connectors can receive a source connector and its options, but this wasn’t possible when using `createAppHook()`. A second parameter has been added, to either pass a specific connector or to change its options:

```js
import connectVoting from '@aragon/connect-voting'
import { createAppHook } from '@aragon/connect-react'

const useVoting = createAppHook(connectVoting, ['thegraph', { subgraphUrl: 'https://…' }])
```

See  f0dbc28de319fdb533a766413de51a3afc564b75.

## createAppHook(): add a dependencies array

Hooks created through `createAppHook()` now accept a dependency array, that can get used to refresh its callback. When no dependency array is passed, the callback never updates.

```js
import connect from '@aragon/connect'
import connectVoting from '@aragon/connect-voting'
import { createAppHook, useApp } from '@aragon/connect-react'

const useVoting = createAppHook(connectVoting)

function App() {
   const [page, setPage] = useState(0)

   const [voting] = useApp('voting')
   const [votes] = useVoting(voting, app => (
      app.votes({ first: 10, skip: 10 * page })
   ), [page])

   return (
      <div>
         <button onClick={() => setPage(page + 1)}>prev</button>
         <button onClick={() => setPage(page - 1)}>next</button>
      </div>
   )
}
```

See 7f661d3231c0769838549be90aa2ce6c60f2b400.

## Populating status.error rather than throwing errors

In many cases, errors were thrown rather than populating the `status.error` object in the React API. It is an issue that was also happening with the JS API (see #229). It has now been fixed.

```js
import connect from '@aragon/connect'
import connectVoting from '@aragon/connect-voting'
import { createAppHook, useApp } from '@aragon/connect-react'

const useVoting = createAppHook(connectVoting)

function App() {

   const [voting, votingStatus] = useApp('voting')
   const [votes, votesStatus] = useVoting(voting, app => app.votes())

   if (votingStatus.error) {
      return (
         <h1>Error: {votingStatus.error.message}</h1>
      )
   }

   if (votesStatus.error) {
      return (
         <h1>Error: {votesStatus.error.message}</h1>
      )
   }

   return (
      // …
   )
}
```

See ddaad5ba02ce582042f13c4e2f86c33150c803b9, 79c538273978f0978496a5b3f95297b54d2510a6.

## App methods subscriptions

Prior to this version, it was not possible to subscribe to app methods using `createAppHook()`. The fetched data was returned once, and only updated when the component was mounted again. The feature has been added, and we can now subscribe to any app methods, alongside the core ones (`useApp()`, `usePermissions()`, …). Every part of the React API is now listening to changes and reacting accordingly.

To use this feature, the subscription methods (`onX`) should be called, but without passing a callback. `createAppHook()` will take care of handling the subscription.

```js
import connect from '@aragon/connect'
import connectVoting from '@aragon/connect-voting'
import { createAppHook, useApp } from '@aragon/connect-react'

const useVoting = createAppHook(connectVoting)

function App() {
   const [voting, votingStatus] = useApp('voting')

   // To enable a subscription, use app.onVotes() rather than app.votes().
   const [votes, votesStatus] = useVoting(voting, app => app.onVotes())

   return (
      // …
   )
}
```

Of course, it also works with the dependencies array, taking care of unsubscribing and subscribing to the new query:

```js
import connect from '@aragon/connect'
import connectVoting from '@aragon/connect-voting'
import { createAppHook, useApp } from '@aragon/connect-react'

const useVoting = createAppHook(connectVoting)

function App() {
   const [page, setPage] = useState(0)

   const [voting] = useApp('voting')
   const [votes] = useVoting(voting, app => (

      // Note the app.onVotes() rather than app.votes().
      app.onVotes({ first: 10, skip: 10 * page })

   // The subscription will refresh every time `page` changes.
   ), [page])

   return (
      // …
   )
}
```

Important: this only works with app connectors enabling this syntax. All the core app connectors have been updated for it, but third party app connectors will require to be updated too.

See #229, 918810bb7ac321b0d1ff76e9f4a4f7d8dbb984c1, 2d2cc88ba83512eb19ddd11492be6f421881c3ee, 02782800d81bbef8ddba8f3d267594ad3706fe51, 65cc59c30323c7e939ea4b2b74b5229be8919e3c, 8e32638eb9405c69428021ed0b0da98049abacb2, 7a173b994d5beae2219c86bbc1639450bd0932e2, 0e77afbe27f3a2346b5dbc9132faf3c29ee6c7a6, 2076bc473076ebd0414a1fedc26be0a0fd435e4c.


packages/connect-react/src/index.ts
```diff
@@ -184,52 +211,108 @@ export function usePermissions(): [Permission[], LoadingStatus] {
...
-    const [result, setResult] = useState<T | null>(null)
-    const [error, setError] = useState<Error | null>(null)
-    const [loading, setLoading] = useState<boolean>(false)
+    const [{ result, error, loading }, setStatus] = useState<{
+      error: Error | null
+      loading: boolean
+      result: T | null
+    }>({
+      error: null,
+      loading: false,
```
andy-hook, line 229: Should the `loading` statuses be defaulting to `true`? at the moment it's confusing because conditionally accessing data behind these flags on the first render will throw on `undefined / null`.

packages/connect-react/src/index.ts
```diff
@@ -146,20 +157,36 @@ function useConnectSubscription<Data>(
...
+        return
       }
+
+      // This is necessary because some connectors might keep providing new app
+      // instances, even if these instances are not actually updated.
+      // For example, the Graph connector uses HTTP polling which has this effect.
+      const dataJson = JSON.stringify({ error, data })
+      if (dataJson === dataJsonRef.current) {
+        return
+      }
```
andy-hook, line 173: This will make controlling updates so much easier ❤️

packages/connect-react/src/index.ts
```diff
@@ -184,52 +211,108 @@ export function usePermissions(): [Permission[], LoadingStatus] {
   return useConnectSubscription<Permission[]>(callback, [])
 }
 
-export function createAppHook(appConnect: Function) {
+export function createAppHook(
+  appConnect: Function,
+  connector?: string | [string, { [key: string]: any } | undefined]
```
andy-hook, line 216: I'm not familiar with the type set-up here but should this connector prop expect `ConnectorDeclaration`?

packages/connect-react/src/index.ts
```diff
@@ -184,52 +211,108 @@ export function usePermissions(): [Permission[], LoadingStatus] {
...
+                    loading: false,
+                    result: error ? null : result,
+                  })
+                }
+              }
+            )
+
+            // Just async data
+          } else {
+            console.log('Subscription mode')
```
andy-hook, line 289: Is this log accurate?

packages/connect-react/src/index.ts
```diff
@@ -146,20 +157,36 @@ function useConnectSubscription<Data>(
       handler?.unsubscribe?.()
```
andy-hook, line 157: Just wanted to check, when a component using a subscription is mounted, is it a new subscription each time? And the unsubs happen in the cleanup like here?

andy-hook (commented):
This is going to be so useful, thanks for turning it around so quickly @bpierre ! I've been doing a little bit of testing and left a few surface level comments. Something I've noticed is with the unsubs, when I unmount `Votes` in the react demo both components seem to be unsubscribed and then loading gets stuck on the remount.

ezgif-7-704b5070f49c

Is this expected?

packages/connect-react/src/index.ts
bpierre: Almost! The exception is that we can’t pass a [`IOrganizationConnector`](https://github.com/aragon/connect/blob/6e16015925133a9649e2c6533f39ead3303789a6/packages/connect/src/connect.ts#L30) here, as it is only meant for the main (organization) connector, not the app connectors that only have a name + a configuration.

packages/connect-react/src/index.ts
bpierre: It’s not, and it shouldn’t have been committed either :smile:

packages/connect-react/src/index.ts
bpierre: > Just wanted to check, when a component using a subscription is mounted, is it a unique subscription?

Yes a new subscription is created for each component using a subscription hook once mounted.

> And the unsubs happen in the cleanup like here?

Yes exactly!

examples/list-votes-react/src/App.tsx
```diff
@@ -7,12 +7,26 @@ import {
 } from '@aragon/connect-react'
 import connectVoting, { Voting, Vote } from '@aragon/connect-voting'
 
+// const NETWORK = 1
+// const ORG_ADDRESS = 'governance.aragonproject.eth'
+// const VOTING_APP_FILTER = 'voting'
```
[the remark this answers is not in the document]
bpierre: Yes!

bpierre:
>  I've noticed is with the unsubs, when I unmount `Votes` in the react demo both components seem to be unsubscribed and then loading gets stuck on the remount. Is this expected?

Good catch, thanks! It should be fixed since add4da4ead869eee868ea66d870a3a5a95353a22. Updates should now be reduced to the strict minimum, and the loading issue is now fixed. Could you please try and tell me if it works for you?

packages/connect-react/src/index.ts
bpierre: Yes you’re right! I updated the default loading states in https://github.com/aragon/connect/pull/233/commits/add4da4ead869eee868ea66d870a3a5a95353a22

andy-hook: Amazing, to use the same example as before: `orgStatus` works great now! but I'm still seeing the same issue with `votesStatus` via `useVoting` though.

andy-hook:
> > I've noticed is with the unsubs, when I unmount `Votes` in the react demo both components seem to be unsubscribed and then loading gets stuck on the remount. Is this expected?
> 
> Good catch, thanks! It should be fixed since [add4da4](https://github.com/aragon/connect/commit/add4da4ead869eee868ea66d870a3a5a95353a22). Updates should now be reduced to the strict minimum, and the loading issue is now fixed. Could you please try and tell me if it works for you?

Awesome stuff, the inclusion of the buttons in the demo was nice too 😁  the mounting issue is fixed! but I still saw the same problem with the `votesStatus` loading flag mentioned here https://github.com/aragon/connect/pull/233#discussion_r476518533

packages/connect-react/src/index.ts
bpierre: @andy-hook Trying to reproduce but I’m not sure how to do it, is the issue that `votes` can be `null`? If that’s the case, this is intended since we don’t know in advance the type of data here, so we can’t really have a default other than `null`. A solution I can see could be to have another optional param to pass an initial value, maybe after the dependencies. If the issue is not this, could you post a way to reproduce it? Thanks!

andy-hook: Sorry if I wasn't clear. I was referring to the `loading` status when using `useVoting`. It reports as `false` initially, even though the votes have yet to be loaded.

bpierre: Oh sorry I misunderstood, it was clear :smile: Let me check for the `loading` status.

Also this gave me the idea of returning `undefined` rather than `null` for the values, so that we can use the default value of the destructuring assignment:

```js
// `votes` is always an array
const [votes = [], votesStatus] = useVoting(voting, app => app.onVotes())
```

bpierre: @andy-hook It should work now, let me know if you still have the issue!

andy-hook:
@bpierre – Loading statuses are consistent and working great! I did spin up the org viewer react demo as well just to double check some things and I'm seeing a `useState` recursion warning in the console. Best way to repro is to spin up that example and take a look, let me know if you see the same thing!

bpierre:
@andy-hook oops https://github.com/aragon/connect/pull/233/commits/ba355742228c627d66912c510923ecd58ef288a6

It should all be fine now, I tested all the examples. Let me know if that works for you!

packages/connect-react/src/index.ts
```diff
@@ -45,12 +49,25 @@ export function Connect({
...
+    orgLoading: true,
+  })
 
   const cancelOrgLoading = useRef<Function | null>(null)
 
+  // If the connector is custom instance, it is the responsibility of the provider to memoize it.
+  const connectorUpdateValue =
+    Array.isArray(connector) || typeof connector === 'string'
+      ? JSON.stringify(connector)
+      : connector
```
andy-hook, line 68: Just wondering with this, what situation might this apply?

andy-hook:
> @andy-hook oops [ba35574](https://github.com/aragon/connect/commit/ba355742228c627d66912c510923ecd58ef288a6)
> 
> It should all be fine now, I tested all the examples. Let me know if that works for you!

Works great!

giphy

packages/connect-react/src/index.ts
bpierre: For example, a third party implementing their own connector (like `@aragon/connector-thegraph`), or just a fork of one of ours if they need to modify anything in it. The `connect()` function also accepts a connector instance if passed as a second parameter.

andy-hook (approved):
This is working great, thank you @bpierre !

@0xGabi you might want to double check the specifics again but it looks good on my end.

PR #234, opened 2020-08-24
CI: Fix typo on GiHub action

0xGabi:


bpierre (approved):

.github/workflows/test.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: yarn --frozen-lockfile
+      - run: yarn prettier
```
bpierre, line 20: Shouldn’t we leave the files as they are, so that `yarn test` can detect formatting issues?

bpierre (commented):
I might be missing something, but shouldn’t we run prettier locally, rather than in a GitHub action? Otherwise it would prevent us to see the formatting issues.

PR #241, opened 2020-08-25
Fix: Add missing SubscriptionCallback types

0xGabi:
Fix CI adding missing SubscriptionCallback + update collateralRequirement tokenId

### DIFF
--- a/packages/connect-voting-disputable/src/thegraph/parsers/collateralRequirements.ts
+++ b/packages/connect-voting-disputable/src/thegraph/parsers/collateralRequirements.ts
@@ -17,7 +17,7 @@ export function parseCollateralRequirement(
     {
       id: collateralRequirement.id,
       voteId: collateralRequirement.vote.id,
-      token: collateralRequirement.token.id,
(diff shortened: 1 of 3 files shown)


packages/connect-voting-disputable/src/thegraph/parsers/collateralRequirements.ts
```diff
@@ -17,7 +17,7 @@ export function parseCollateralRequirement(
     {
       id: collateralRequirement.id,
       voteId: collateralRequirement.vote.id,
-      token: collateralRequirement.token.id,
+      tokenId: collateralRequirement.token.id,
```
bpierre, line 20: Should we update anything else related to this change? :thinking:

bpierre (approved):
LGTM!

packages/connect-voting-disputable/src/thegraph/parsers/collateralRequirements.ts
0xGabi: I double-check and no we don't. It was a change introduced during a merge conflict.

PR #254, opened 2020-08-30
Fix precedence campaign subgraphs

facuspagnuolo:
Using a triple equal for this wasn't working 🤷

### DIFF
--- a/packages/connect-agreement/subgraph/src/aragon-hooks.ts
+++ b/packages/connect-agreement/subgraph/src/aragon-hooks.ts
@@ -13,7 +13,7 @@ export function getTemplateForApp(appId: string): string | null {
   const AGREEMENT_OPEN = '0x34c62f3aec3073826f39c2c35e9a1297d9dbf3cc77472283106f09eee9cf47bf'
   const AGREEMENT_PRECEDENCE_CAMPAIGN = '0x15a969a0e134d745b604fb43f699bb5c146424792084c198d53050c4d08126d1'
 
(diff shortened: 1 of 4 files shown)


packages/connect-agreement/subgraph/src/aragon-hooks.ts
```diff
@@ -13,7 +13,7 @@ export function getTemplateForApp(appId: string): string | null {
   const AGREEMENT_OPEN = '0x34c62f3aec3073826f39c2c35e9a1297d9dbf3cc77472283106f09eee9cf47bf'
   const AGREEMENT_PRECEDENCE_CAMPAIGN = '0x15a969a0e134d745b604fb43f699bb5c146424792084c198d53050c4d08126d1'
 
-  if (appId === AGREEMENT_OPEN || appId === AGREEMENT_PRECEDENCE_CAMPAIGN) {
+  if (appId == AGREEMENT_OPEN || appId == AGREEMENT_PRECEDENCE_CAMPAIGN) {
```
bpierre, line 16: Ah, it probably means that `appId` is set to a number rather than a string.

Instead of relying on implicit conversion, should we either do it manually (using `String(appId)`) or use numbers for `AGREEMENT_OPEN` and `AGREEMENT_PRECEDENCE_CAMPAIGN` (by removing the quotes)?

Just thinking that we might want to prevent any use of `==` with ESLint at some point.

bpierre (approved):

PR #258, opened 2020-09-01
DisputableVoting: Implement vote settled at date

facuspagnuolo:


packages/connect-disputable-voting/src/thegraph/queries/index.ts
```diff
@@ -112,6 +112,7 @@ export const GET_VOTE = (type: string) => gql`
       quietEndingExtensionDuration
       quietEndingSnapshotSupport
       script
+  settledAt
```
0xGabi, line 115:
```suggestion
      settledAt

PR #263, opened 2020-09-05
DisputableVoting: Implement more states

facuspagnuolo:
This PR includes:
- Return votes ordered by start date
- Add "disputed at" date
- Support "settled" status for votes
- Implement artbirator fees 
- Support quiet ending for vote end date computation


packages/connect-disputable-voting/src/__test__/thegraph/votes.test.ts
```diff
@@ -72,14 +72,14 @@ describe('DisputableVoting votes', () => {
     })
 
     test('allows fetching a single vote', () => {
-      const vote = votes[6]
+      const vote = votes[2]
```
bpierre, line 75: Should we do [something like this](https://github.com/aragon/connect/commit/fd67970e75f462cde9858e10a5ba719f4f6fe065) to prevent this to break when votes get added?

packages/connect-disputable-voting/src/models/ArbitratorFee.ts
```diff
@@ -0,0 +1,40 @@
...
+    return formatBn(this.amount, this.tokenDecimals)
+  }
+
+  async token(): Promise<ERC20> {
+    return this.#connector.ERC20(this.tokenId)
+  }
+
+  onToken(callback: Function): SubscriptionHandler {
+    return this.#connector.onERC20(this.tokenId, callback)
+  }
```
bpierre, line 39: If we want this method to be supported by `createAppHook()` in the React API, we should do this instead:

```suggestion
  onToken(callback?: SubscriptionCallback<ERC20>): SubscriptionResult<ERC20> {
    return subscription<ERC20>(callback, (callback) =>
      this.#connector.onERC20(this.tokenId, callback)
    )
  }
```

You can find an example here: https://github.com/aragon/connect/blob/3bef799a65a985fbd63949f7fe3f5b852130d5e1/packages/connect-tokens/src/models/Tokens.ts#L37-L44

packages/connect-disputable-voting/src/models/Vote.ts
```diff
@@ -75,23 +81,35 @@ export default class Vote {
...
   get endDate(): string {
-    const originalEndDate = bn(this.startDate).add(bn(this.duration))
-    const endDateAfterPause = originalEndDate.add(bn(this.pauseDuration))
-    return endDateAfterPause.add(bn(this.quietEndingExtensionDuration)).toString()
+    const baseVoteEndDate = bn(this.startDate).add(bn(this.duration))
+    const endDateAfterPause = baseVoteEndDate.add(bn(this.pauseDuration))
+    const lastComputedEndDate = endDateAfterPause.add(bn(this.quietEndingExtensionDuration))
+
+    // The last computed end date is correct if we have not passed it yet or if no flip was detected in the last extension
+    const currentTimestamp = bn(parseInt((Date.now() / 1000).toString()).toString())
```
bpierre, line 106: No need to change anything, but I was wondering if there was reason to do `toString()` => `parseInt()` => `toString()`? Or if we could use `String(Math.floor(Date.now() / 1000))`?

packages/connect-disputable-voting/src/thegraph/connector.ts
```diff
@@ -285,6 +287,26 @@ export default class DisputableVotingConnectorTheGraph
...
+      { arbitratorFeeId },
+      (result: QueryResult) => parseArbitratorFee(result, this)
+    )
+  }
+
+  onArbitratorFee(
+    arbitratorFeeId: string,
+    callback: SubscriptionCallback<ArbitratorFee | null>
+  ): SubscriptionHandler {
+    return this.#gql.subscribeToQueryWithParser<ArbitratorFee | null>(
```
bpierre, line 302: Same than above, we could use `subscription()` / `SubscriptionResult` here.

bpierre (approved):
Looking good! :fire::fire::fire: I left a few comments, mostly for React compatibility.

packages/connect-disputable-voting/src/__test__/thegraph/votes.test.ts
facuspagnuolo: Interesting, note that in this PR introduced a sorting instruction for the query, then votes should be ordered, but I see your point definitely

packages/connect-disputable-voting/src/models/Vote.ts
facuspagnuolo: Just a brainfart haha, we can do that for sure :)

packages/connect-disputable-voting/src/thegraph/connector.ts
facuspagnuolo: Compared to other apps it seems like we are returning `SubscriptionHandler`s from the connectors always, is that correct?

bpierre: Yeah we might need to update other parts: `SubscriptionHandler` is now more of an internal type used by `SubscriptionResult`, which is the one we should use from subscription methods :-)

PR #256, opened 2020-08-31
Refactor API: ForwardingPath descriptions + App intent actions

0xGabi:
This PR includes most of the changes discussed in #132 

One of the most relevant changes from the spec is that we don't have an Intent object as it was an intermediary step that ends up adding more complexity to the code. Instead of giving an intent we calculate the path and then create a `ForwardingPath` object that provide extra functionality on top of the data:
- Sign transactions with a signer
- Describe the path using Radspec
- Apply pre-transactions to the path

PR #264, opened 2020-09-06
Agreement: Implement actions

facuspagnuolo:
**MERGE AFTER #256**

This PR introduces the following agreement actions
- sign
- challenge
- dispute
- settle
- close


packages/connect-agreement/src/models/Agreement.ts
```diff
@@ -147,7 +153,69 @@ export default class Agreement {
...
+  }
+
+  sign(signerAddress: string, versionNumber: string): Promise<ForwardingPath> {
+    return this.#app.intent('sign', [versionNumber], { actAs: signerAddress })
+  }
+
+  async challenge(actionNumber: string, settlementOffer: string, finishedEvidence: boolean, context: string, signerAddress: string): Promise<ForwardingPath> {
+    const intent = await this.#app.intent('challengeAction', [actionNumber, settlementOffer, finishedEvidence, utils.toUtf8Bytes(context)], { actAs: signerAddress })
+
+    const action = (await this.action(actionNumber))!
```
bpierre, line 181: Instead of using `!`: should we test for it to be `null`, throw an error if that’s the case, then use `as Action`?

packages/connect-agreement/src/__test__/models/agreement.test.ts
```diff
@@ -237,4 +210,126 @@ describe('Agreement', () => {
       })
     })
   })
+
+  describe('sign', () => {
```
0xGabi, line 214: Nice!

packages/connect-agreement/src/__test__/models/agreement.test.ts
```diff
@@ -237,4 +210,126 @@ describe('Agreement', () => {
...
+    const VERSION_NUMBER = '1'
+    const SIGNER_ADDRESS = '0x0090aed150056316e37fe6dfa10dc63e79d173b6'
+
+    it('returns a sign intent', async () => {
+      const abi = new ethers.utils.Interface(['function sign(uint256)'])
+      const intent = await agreement.sign(SIGNER_ADDRESS, VERSION_NUMBER)
```
0xGabi, line 220: One thing that I'm still a bit unsure is about calling the return value `intent`. As I understood the intent is only the action itself we are trying to do. The return value of intent is a `TransactionPath`/`ForwardingPath`.

packages/connect-agreement/src/models/Agreement.ts
```diff
@@ -69,36 +71,40 @@ export default class Agreement {
     )
   }
 
+  async disputeFees(versionId: string): Promise<any> {
+    const version = await this.#connector.version(versionId)
+    const arbitrator = new Contract(version.arbitrator, arbitratorAbi, this.#app.provider)
```
0xGabi, line 76: Nice, I think this is quite clean. The only thing is the `arbitratorAbi` that we can eventually consider to move it to a Disputables utility package.

packages/connect-agreement/src/models/Agreement.ts
```diff
@@ -147,7 +153,77 @@ export default class Agreement {
...
+    return this.#connector.action(this.actionId(actionNumber))
+  }
+
+  onAction(acitonNumber: string, callback?: SubscriptionCallback<Action | null>): SubscriptionResult<Action | null> {
+    return subscription<Action | null>(callback, (callback) =>
+      this.#connector.onAction(this.actionId(acitonNumber), callback)
     )
   }
+
+  sign(signerAddress: string, versionNumber: string): Promise<ForwardingPath> {
```
0xGabi, line 182: I wonder if the `signerAddress` should be optional and part of a similar `PathOptions` object like we define here: https://github.com/aragon/connect/issues/132.

We can try to get an account from the provider or otherwise error if none was provided as an option?

packages/connect-agreement/src/__test__/models/agreement.test.ts
facuspagnuolo: Not sure about that too tbh

packages/connect-core/src/utils/path/calculatePath.ts
```diff
@@ -45,15 +45,13 @@ async function calculateForwardingPath(
...
+      const forwardingFeePreTransactions = await buildForwardingFeePreTransactions(
         transaction,
         provider
       )
       // If that happens, we give up as we should've been able to perform the action with this
       // forwarding path
-      return forwardingFeePretransaction
-        ? [forwardingFeePretransaction, transaction, ...path]
-        : [transaction, ...path]
+      return [...forwardingFeePreTransactions, transaction, ...path]
```
0xGabi, line 54: 😎

packages/connect-core/src/utils/transactions.ts
```diff
@@ -60,55 +60,65 @@ export function createForwarderTransactionBuilder(
...
 
-export async function buildApprovePretransaction(
+export async function buildApprovePreTransactions(
   transaction: Transaction,
   tokenData: TokenData,
   provider: ethersProviders.Provider
-): Promise<Transaction | undefined> {
-  // Token allowance pretransactionn
+): Promise<Transaction[]> {
+  // Token allowance pre-transactionn
```
0xGabi, line 68:
```suggestion
  // Token allowance pre-transaction

0xGabi (approved):
Solid,  the `buildApprovePreTransactions` looks great!

0xGabi:
My bad, I should have publish the new app version from this branch instead of merging it to the refactor branch. I'll do that with the `Disputable Voting`

PR #265, opened 2020-09-07
Typed errors

bpierre:
Fixes https://github.com/aragon/connect/pull/265

This PR adds typed errors, that are exposed to let consumers know more precisely why something failed.


docs/connectors/finance-app.md
```diff
@@ -4,17 +4,36 @@ This is an app connector for the Finance app (`finance.aragonpm.eth`). It only s
...
+
+It can throw the following errors:
+
+| Error type                                                     | Description                                                                                                                      |
+| -------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| [`ErrorInvalidApp`](./errors.md#error-invalid-app)             | Either the passed value is not a valid app object, or its name is not `finance` (`err.reason === "wrong-name"`).                 |
+| [`ErrorInvalidConnector`](./errors.md#error-invalid-connector) | Either the connector configuration format is not valid, or the connector name is not supported (`err.reason === "unsupported"`). |
+| [`ErrorInvalidConnector`](./errors.md#error-invalid-connector) | The connector is not supported.                                                                                                  |
```
0xGabi, line 35: Is there a difference with the above error?

packages/connect-core/src/errors.ts
```diff
@@ -0,0 +1,168 @@
+function formatMessage(message: string, reason: string): string {
+  return message + reason ? ` (reason: ${reason})` : ''
+}
+
+function defineNonEnumerable(instance: object, name: string, value: any) {
+  Object.defineProperty(instance, name, { value, enumerable: false })
```
0xGabi, line 6: This is interesting, I had to read the documentation on `Object.defineProperty` to better understand what you were doing here. I notice the `enumerable` default value is `false. But I guess better to be explicit.

packages/connect-thegraph/src/connector.ts
```diff
@@ -101,19 +104,31 @@ class ConnectorTheGraph implements IOrganizationConnector {
...
-    )
+    try {
+      return this.#gql.performQueryWithParser<Role[]>(
+        queries.ROLE_BY_APP_ADDRESS('query'),
+        { appAddress: appAddress.toLowerCase() },
+        async (result) => parseRoles(result, organization)
+      )
+    } catch (err) {
+      throw new ErrorUnexpectedResult(
+        'Unexpected result when fetching the roles.'
```
0xGabi, line 115: I think we should pass the `err` as the `reason` option otherwise we swallow the errors from the `GraphQLWrapper` don't we?

0xGabi (approved):
Reviewing these changes was super interesting to learn how to properly handle Errors in a library. @bpierre you really nailed it with the Error types you choose. Is going to be super useful for the consumers to better understand the context of the problems they may face.

🔌❤️

packages/connect-core/src/errors.ts
bpierre: Ah yes, this is all so that these properties don’t appear in the console when an error is thrown :smile:

packages/connect-thegraph/src/connector.ts
bpierre: Yes this could definitely be an issue.  I think it could be a problem if we start using `reason` for that, as it is supposed to be a string that helps to distinguish between two errors of the same type. We might also have a chain of errors, I wonder if it would be useful to provide a way to users so they can explore it? Or if we should let this to the debugger. :thinking:

I was discussing with @sohkai about `reason`, and he was making the point that it shouldn’t be that useful, as apps probably won’t act differently based on the `reason`, but on the type only. For example, if an `ErrorConnection` fails for X or Y reason, there shouldn’t be a need for the app to do anything else than retrying to connect. Authors will also have the error message that will be specific and provide context. If making the distinction between two errors of the same type really matters, then it might be a sign that we need to extend the error (which is more or less the role `reason` is having anyway, by acting as a “light subtype”).

Now for this specific case: I think we could suggest to enable the verbosity in the message itself, so authors can see why the query failed and debug it? Something related: I would like to discuss the possibility to have different levels of verbosity rather than a global verbosity flag, having this would also help here.

What do you think?

docs/connectors/finance-app.md
bpierre: Technically yes, it could be thrown from either [`createAppConnector()`](https://github.com/aragon/connect/blob/errors/packages/connect-core/src/utils/app-connectors.ts#L55-L59) directly, or [from the app connector](https://github.com/aragon/connect/blob/errors/packages/connect-finance/src/connect.ts#L19-L23) implementation.

But your comment made me realize this distinction doesn’t really matter for consumers, I removed it :smile:

packages/connect-thegraph/src/connector.ts
bpierre: Update: I removed `reason`.

PR #269, opened 2020-09-07
Docs: add createAppHook()

bpierre:


docs/guides/connect-with-react.md
```diff
@@ -108,3 +108,118 @@ This component is required in order to use the provided hooks.
 | Name    | Type                                                                          | Description                                                                                                     |
 | ------- | ----------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
 | returns | `[Permission[], { loading: boolean, error: null \| Error, retry: Function }]` | An array containing the organization [permissions](../api-reference/permission.md) and a loading status object. |
+
+### createAppHook()
```
0xGabi, line 112: Wonder if we want to give a bit more context about the return type and description like above?

I think this is especially important as is one of the only places that we have a curry function.

0xGabi (commented):
Solid! 💪 Left a small suggestion.

docs/guides/connect-with-react.md
bpierre: Ah yes good idea, I will add this. :+1:

PR #283, opened 2020-09-10
DisputableVoting: Fix collateral requirements parsing

facuspagnuolo:


0xGabi (approved):
You are one step ahead! 💪

I was wondering yesterday during the debug session why we didn't use the `loadOrCreate` pattern here.

PR #286, opened 2020-09-11
Agreement: Reduce staking movements complexity

facuspagnuolo:
@0xGabi I may need some help here, I set up a separate DAO with a separate subgraph for the tests but the organizations subgraph is not retrieving it


packages/connect-agreement/src/__test__/models/agreement.test.ts
```diff
@@ -1,15 +1,16 @@
...
-const AGREEMENT_APP_ADDRESS = '0x9c92dbd8a8e5903e2741202321073091109f26be'
-const AGREEMENT_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/aragon/aragon-agreement-rinkeby-staging'
+const ORGANIZATION_NAME = '0x6322eb0294c6aadb7e1b37d41fd605a34df661dc'
+const AGREEMENT_APP_ADDRESS = '0xe4575381f0c96f58bd93be6978cc0d9638d874a2'
+const AGREEMENT_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/facuspagnuolo/aragon-agreement-rinkeby-staging'
 
-describe('Agreement', () => {
+
+// TODO: for some reason this DAO is not retrieved from the organizations subgraph
+describe.skip('Agreement', () => {
```
facuspagnuolo, line 13: @0xGabi here ☝️

packages/connect-agreement/subgraph/src/Agreement.ts
```diff
@@ -294,6 +270,91 @@ function updateCollateralRequirement(agreement: Address, disputable: Address, co
...
+  if (type == 'new') {
+    movement.amount = collateralAmount
+    movement.actionState = 'Scheduled'
+    movement.collateralState = 'Locked'
+  } else if (type == 'challenged') {
+    movement.amount = collateralAmount
+    movement.actionState = 'Challenged'
+    movement.collateralState = 'Challenged'
+    staking.challenged = staking.challenged.plus(collateralAmount)
+  } else if (type == 'settled') {
```
0xGabi, line 306: I wonder if we need to check that the `collateralAmount` be less than the `staking.challenged`?

0xGabi (approved):
Great work reducing the complexity of the staking movements. I think it make sense to be part of the Agreement mapping 👍

PR #287, opened 2020-09-12
DisputableVoting: Implement voting checks

facuspagnuolo:
Implemented disputable voting checks:
- can vote
- can execute
- has voted
- current balance
- snapshot balance


packages/connect-disputable-voting/src/models/ERC20.ts
```diff
@@ -1,15 +1,34 @@
+import { miniMeAbi } from '@aragon/connect-core'
+import { BigNumber, Contract, providers as ethersProviders } from 'ethers'
+
 import { ERC20Data } from '../types'
 
 export default class ERC20 {
+  #provider: ethersProviders.Provider
```
bpierre, line 7: Should we name this `ethersProvider`? This is because we also have Ethereum providers (EIP-1193) in the library.

packages/connect-disputable-voting/src/models/ERC20.ts
```diff
@@ -1,15 +1,34 @@
...
+
     this.id = data.id
     this.name = data.name
     this.symbol = data.symbol
     this.decimals = data.decimals
   }
+
+  async getBalance(address: string): Promise<BigNumber> {
+    return this._contract().balanceOf(address)
+  }
```
bpierre, line 25: Should we name this method `balance()`? Not for this PR, but I am also thinking that it could be nice to have a `onBalance()` here (by querying it on the block event).

packages/connect-disputable-voting/src/models/Vote.ts
```diff
@@ -172,6 +177,46 @@ export default class Vote {
...
+
+  async token(): Promise<ERC20> {
+    return this.#connector.ERC20(this.tokenId)
+  }
+
+  async hasVoted(voterAddress: Address): Promise<boolean> {
+    const castVote = await this.castVote(voterAddress)
+    return castVote !== null
+  }
+
```
bpierre, line 219: Not necessarily for this PR, but we should probably have subscription equivalents for these methods, apart from `token()`.

packages/connect-disputable-voting/src/thegraph/connector.ts
```diff
@@ -48,20 +46,23 @@ type DisputableVotingConnectorTheGraphConfig = {
   verbose?: boolean
 }
 
-export default class DisputableVotingConnectorTheGraph
-  implements IDisputableVotingConnector {
+export default class DisputableVotingConnectorTheGraph implements IDisputableVotingConnector {
   #gql: GraphQLWrapper
+  #provider: ethersProviders.Provider
```
bpierre, line 51: Here too, I think using `#ethersProvider` would improve clarity.

bpierre (approved):
LGTM! I added some comments about naming and subscriptions.

packages/connect-disputable-voting/src/models/Vote.ts
facuspagnuolo: I'm starting to think we should find a more automated way to have these haha

bpierre: Definitely 😬

packages/connect-disputable-voting/src/models/Vote.ts
```diff
@@ -172,6 +177,46 @@ export default class Vote {
...
+    return token.getBalanceAt(voterAddress, this.snapshotBlock)
+  }
+
+  async formattedVotingPower(voterAddress: Address): Promise<string> {
+    const token = await this.token()
+    const balance = await token.getBalanceAt(voterAddress, this.snapshotBlock)
+    return formatBn(balance, token.decimals)
+  }
+
+  async token(): Promise<ERC20> {
```
0xGabi, line 211: I wonder if the `ERC20` model is generic enough to have a handler at the organization to construct it. One advantage with that would be to prevent doing a subgraph query every time we want to use the token.

0xGabi (approved):
LGTM!

packages/connect-disputable-voting/src/models/Vote.ts
bpierre: Agreed, we could at least provide a base model for connectors to extend. :+1:

PR #291, opened 2020-09-15
DisputableVoting: Implement settlement offer

facuspagnuolo:


packages/connect-disputable-voting/src/models/Vote.ts
```diff
@@ -295,6 +297,16 @@ export default class Vote {
...
   }
 
+  async formattedSettlementOffer(): Promise<string | null> {
+    if (!this.settlementOffer) {
+      return null
+    }
+
+    const collateralRequirement = await this.collateralRequirement()
+    const token = await collateralRequirement.token()
+    return formatBn(this.settlementOffer, token.decimals)
```
0xGabi, line 307: Probably something to improve going forward but I think that doing two subgraph requests to get the `token.decimals` is going to become expensive when the graph has an indexer network.

0xGabi (approved):
LGTM

packages/connect-disputable-voting/src/models/Vote.ts
facuspagnuolo: definitely, we can improve this

PR #292, opened 2020-09-16
Core: prefer generating signature from ABI

0xGabi:
Reuse changes from https://github.com/aragon/aragon-cli/pull/1687

This fix prevents using the wrong signature hash from artifact functions by preferring creating the signature from the ABI definition.

### DIFF
--- a/packages/connect-core/src/utils/app.ts
+++ b/packages/connect-core/src/utils/app.ts
@@ -1,10 +1,31 @@
 import { utils as ethersUtils } from 'ethers'
 
-import { AppMethod } from '../types'
+import { Abi, AppMethod } from '../types'
 import App from '../entities/App'
 
 export const apmAppId = (appName: string): string =>
   ethersUtils.namehash(`${appName}.aragonpm.eth`)
+
+const checkSignature = (signature: string, abi: Abi): string => {
+  const regex = signature.match(/(.*)\((.*)\)/m)!
+
+  const name = regex[1]
+  const params = regex[2].split(',')
+
+  // If a single ABI node is found with function name and same number of parameters,
+  // generate the signature from ABI. Otherwise, use the one from artifact.
+  const functionAbis = abi
+    .filter((node) => node.name === name)
+    .filter((node) => node.inputs.length === params.length)
+
+  if (functionAbis.length === 1) {
+    return `${functionAbis[0].name}(${functionAbis[0].inputs.map(
+      (input) => input.type
+    )})`
+  }
+
+  return signature
+}
 
 function findAppMethod(
   app: App,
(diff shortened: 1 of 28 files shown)


packages/connect-core/src/utils/app.ts
```diff
@@ -1,11 +1,32 @@
...
 
-import { AppMethod } from '../types'
+import { Abi, AppMethod } from '../types'
 import App from '../entities/App'
 
 export const apmAppId = (appName: string): string =>
   ethersUtils.namehash(`${appName}.aragonpm.eth`)
 
+const checkSignature = (signature: string, abi: Abi): string => {
+  const regex = signature.match(/(.*)\((.*)\)/m)!
```
bpierre, line 10: Should we remove the `!` and test ourselves if the regex matches? I think we should remove any `!` remaining in the codebase as it makes the code using it unsafe.

Also should we call the variable `matches` (since the regex is on the right side here)?

packages/connect-core/src/utils/app.ts
```diff
@@ -1,11 +1,32 @@
 import { utils as ethersUtils } from 'ethers'
 
-import { AppMethod } from '../types'
+import { Abi, AppMethod } from '../types'
 import App from '../entities/App'
 
 export const apmAppId = (appName: string): string =>
   ethersUtils.namehash(`${appName}.aragonpm.eth`)
 
+const checkSignature = (signature: string, abi: Abi): string => {
```
bpierre, line 9: I would expect `checkSignature()` to return a boolean, should it be something like `signatureFromAbi()` instead? :thinking: 

Also we should use `function checkSignature() {` to follow the coding style conventions.

packages/connect-core/src/utils/app.ts
```diff
@@ -1,11 +1,32 @@
...
+  // If a single ABI node is found with function name and same number of parameters,
+  // generate the signature from ABI. Otherwise, use the one from artifact.
+  const functionAbis = abi
+    .filter((node) => node.name === name)
+    .filter((node) => node.inputs.length === params.length)
+
+  if (functionAbis.length === 1) {
+    return `${functionAbis[0].name}(${functionAbis[0].inputs.map(
+      (input) => input.type
+    )})`
```
bpierre, line 24: What would you think of making the comma explicit? It took me a while to remember it was the default of `Array#toString()` haha
 
```suggestion
    return `${functionAbis[0].name}(${functionAbis[0].inputs.map(
      (input) => input.type
    ).join(',')})`
```

bpierre (approved):
LGTM, I only left a few comments about a `!` and naming / conventions.

PR #296, opened 2020-09-21
IPFS: use the resolver everywhere + caching mechanism

bpierre:
Also uses unfetch rather than relying on Ethers for fetch() calls.


packages/connect-core/src/utils/metadata.ts
```diff
@@ -19,22 +20,26 @@ export function parseMetadata(name: string, metadata: string): any {
...
+  }
+
+  const url = `${DEFAULT_IPFS_GATEWAY}/ipfs/${contentHash}/${fileName}`
+  try {
+    // await before returning to catch any error in the callback
+    return await metadataCacheStore.get(url, () =>
+      fetch(url).then((res) => res.json())
+    )
+  } catch (error) {
+    throw new ErrorConnection(`Couldn’t fetch ${url}, failed with error.`)
```
0xGabi, line 41: Here we don't want to provide further details on the error other than it is a Connection related error?

packages/connect-core/src/utils/misc.ts
```diff
@@ -1,3 +1,54 @@
...
+
+  return {
+    clear(): void {
+      cachedEntries.length = 0
+    },
+    cachedIndex(id: string): number {
+      return cachedEntries.findIndex((entry) => entry && id === entry[0])
+    },
+    touch(index: number): void {
+      cachedEntries.unshift(cachedEntries.splice(index, 1)[0])
```
0xGabi, line 29: Here we are using `splice` because we want to mutate the cache and remove the element from `index` and place it first in the array?

packages/connect-core/src/utils/misc.ts
```diff
@@ -1,3 +1,54 @@
...
+  touch: (index: number) => void
+} {
+  const cachedEntries: [string, T][] = Array(limit)
+
+  return {
+    clear(): void {
+      cachedEntries.length = 0
+    },
+    cachedIndex(id: string): number {
+      return cachedEntries.findIndex((entry) => entry && id === entry[0])
```
0xGabi, line 26: I wonder if it's a good opportunity to prevent using `&&` and instead use `??`

packages/connect-core/src/utils/misc.ts
```diff
@@ -1,3 +1,54 @@
...
+
+export function createCacheStore<T = string>(
+  limit: number = 10
+): {
+  cachedIndex: (id: string) => number
+  clear: () => void
+  get: (id: string, callback: CacheStoreCallback<T>) => Promise<T>
+  touch: (index: number) => void
+} {
+  const cachedEntries: [string, T][] = Array(limit)
```
0xGabi, line 19: Given the type definition, the `cacheEntries` is an array of tuples with the first element a string and the second element of type T. That's why we use `entry[0]` on line 256 and we return the `cachedEntries[cachedIndex][1]`?

packages/connect-core/src/utils/misc.ts
```diff
@@ -1,3 +1,54 @@
...
+      // Prevents to cache the same value multiple times, in case the same
+      // value gets loaded several times in parallel.
+      cachedIndex = this.cachedIndex(id)
+      if (cachedIndex > -1) {
+        this.touch(cachedIndex)
+        return cachedEntries[cachedIndex][1]
+      }
+
+      cachedEntries.unshift([id, data])
+      cachedEntries.length = limit
```
0xGabi, line 49: Why we are only interested in the final length of the cache? We update on every `get` request in case it was `clear` in between?

0xGabi (approved):
Looking great, left a few comments mostly to better understand the work you did 🤓

packages/connect-core/src/utils/misc.ts
bpierre: Exactly yes, just a trick to move `index` at the first position :-)

packages/connect-core/src/utils/misc.ts
bpierre: Yes! We could even check for `undefined` to be super specific since it’s what we want to test here.

packages/connect-core/src/utils/misc.ts
bpierre: This is for when we start having more than `limit` entries, to ensure the cached entries never exceed `limit`.

packages/connect-core/src/utils/misc.ts
bpierre: Yes that’s correct!

packages/connect-core/src/utils/metadata.ts
bpierre: Are you thinking of having a different types of error for the actual fetching, the JSON parsing, and the cache store?

0xGabi: No, I think the Connection type is fine. My question was to have a better understanding of when we should display the `error` message to give further details. And when we can mute the message.

packages/connect-core/src/utils/misc.ts
bpierre: Thinking more about it, the value can only be either an array or `undefined`, so it shouldn’t be a problem. Still changing it to `??` as it is the new `||` :smile:

packages/connect-core/src/utils/metadata.ts
bpierre: I made it throw two different types of error: https://github.com/aragon/connect/pull/296/commits/444113d59bfd858cc78251a1562340cb6d646723

bpierre:
@0xGabi Could you please give it one last look? I added a few more things :eyes:

0xGabi (approved):
LGTM! 👍

PR #259, opened 2020-09-02
Subgraph: Update main subgraph + Voting subgraph and connector (#259) 

0xGabi:
Close #164, #153

This PR includes:
- default system apps information
- a new `createdAt` value for the `Organization` entity
- removes metada handlers relaying on IPFS
- implements the `loadOrCreate` pattern for entities

PR #301, opened 2020-11-11
Voting: update subgraph + connector

0xGabi:
This PR does:
- Fix an issue with count of cast votes. 
- Update subgraph schema to use string for the `script` argument. Using bytes has a limitation on the size of scripts that could be store on graph-node.
- Update the voting connector with the changes
- Bump connector version: `@aragon/connect-voting@0.7.1
- Add pagination parameters to the casts


bpierre (approved):
LGTM! Should we update the Voting docs before merging it?

0xGabi:
Merged on top of #259 + update docs.

PR #336, opened 2021-11-07
Voting connector: Support statuses

PJColombo:


packages/connect-voting/subgraph/src/Voting.ts
```diff
@@ -118,6 +119,16 @@ export function updateVoteState(votingAddress: Address, voteId: BigInt): void {
   const vote = VoteEntity.load(buildVoteEntityId(votingAddress, voteId))!
   vote.yea = voteData.value6
   vote.nay = voteData.value7
+  vote.isAccepted = isAccepted(vote.yea, vote.nay, vote.votingPower, vote.supportRequiredPct, vote.minAcceptQuorum, votingApp.PCT_BASE())
 
   vote.save()
 }
+
+function isAccepted(yeas: BigInt, nays: BigInt, votingPower: BigInt, supportRequiredPct: BigInt, minimumAcceptanceQuorumPct: BigInt, pctBase: BigInt): boolean {
+  return hasReachedValuePct(yeas, yeas.plus(nays), supportRequiredPct, pctBase) &&
```
0xGabi, line 128:
```suggestion
  return hasReachedValuePct(yeas, yeas.plus(nays).abs(), supportRequiredPct, pctBase) &&

0xGabi (changes requested):
Well done! I left a minor suggestion.

I wonder if you consider using an enum for the status value instead of returning the string. What do you think of touching base with DeepDAO to know their preference about how to consume the status data?

PJColombo:
> Well done! I left a minor suggestion.
> 
> I wonder if you consider using an enum for the status value instead of returning the string. What do you think of touching base with DeepDAO to know their preference about how to consume the status data?

Yes, an enum would be better

packages/connect-voting/subgraph/src/Voting.ts
PJColombo: Hey @0xGabi. What's the rationale behind using `abs()`?. We're always adding two positive numbers here.

0xGabi: Forgot what I said. I got confused 🙃

0xGabi (approved):
LGTM

PR #338, opened 2021-11-12
Voting Connector: Support vote rewards

PJColombo:
This PR builds on top of #336 and addreses the [DeepDao's Gitcoin bounty](https://gitcoin.co/issue/deepdao1/global-dao-hackaton-10.2021/3/100026940) related to supporting vote rewards.

It adds a new functionality to the `Vote` model that allows you to get all the actions inside the vote  `script` which is a set of codified DAO's app contract calls that are executed once the vote is enacted.

If an action involves a token transfer from the DAO to an external account then a `rewards` field is filled.

The following are considered to be  token transfer actions:

- Calls to the Vault/Agent's `transfer(address _token, address _to, uint256 _value)` function
- Calls to the Finance's `newImmediatePayment(address _token, address _receiver, uint256 _amount, string _reference)`

### DIFF
--- a/packages/connect-core/src/utils/app.ts
+++ b/packages/connect-core/src/utils/app.ts
@@ -1,4 +1,4 @@
-import { utils as ethersUtils } from 'ethers'
+import { utils as ethersUtils, utils } from 'ethers'
 
 import { Abi, AppMethod } from '../types'
 import { ErrorInvalid } from '../errors'
@@ -8,6 +8,10 @@ export const apmAppId = (appName: string): string =>
   ethersUtils.namehash(`${appName}.aragonpm.eth`)
 
 function signatureFromAbi(signature: string, abi: Abi): string {
+  if (signature === 'fallback') {
+    return '()'
+  }
+
   const matches = signature.match(/(.*)\((.*)\)/m)
 
   if (!matches) {
@@ -44,7 +48,7 @@ function findAppMethod(
   if (Array.isArray(functions)) {
     method = functions
       .map((f) => {
-        return { ...f, sig: signatureFromAbi(f.sig, app.abi) }
+        return { ...f, sig: signatureFromAbi(f.sig, app.abi), params: [] }
       })
       .find(methodTestFn)
   }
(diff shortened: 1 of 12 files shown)


packages/connect-voting/src/helpers/actions.ts
```diff
@@ -0,0 +1,46 @@
+import { utils } from 'ethers'
+import { AppMethod } from "@aragon/connect"
+import { Reward } from '../types'
+
+export const getRewards = (appId: string, fnData: AppMethod): Reward[] => {
```
0xGabi, line 5: What was your final thought about returning a `Reward` array here? I can't think in which cases we will have multiple rewards on a single action.

0xGabi (approved):
Great work and well done with the unit testing. Reading them made me realize that the final structure you choose for the data was the right one. I left a couple of comments and an open question about your thoughts for the `getRewards` function.

packages/connect-core/src/utils/app.ts
```diff
@@ -8,6 +8,10 @@ export const apmAppId = (appName: string): string =>
   ethersUtils.namehash(`${appName}.aragonpm.eth`)
 
 function signatureFromAbi(signature: string, abi: Abi): string {
+  if (signature === 'fallback') {
+    return '()'
```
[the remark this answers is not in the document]
0xGabi: What do you think of returning the whole signature here? I did a similar fix on the 1Hive fork and I remember we had some issues with radspec descriptions otherwise.

packages/connect-voting/src/helpers/actions.ts
PJColombo: I was thinking about the possibility of having ac action that calls a contract function that receives a list of recipient addresses and transfer tokens to each of them.

0xGabi: That's a good point! Thanks for clarifying it.

pmlambert:
> 

@0xGabi I spent a nice chunk of time not knowing this 🥇

packages/connect-voting/src/__test__/votes.test.ts
```diff
@@ -1,11 +1,19 @@
+import { BigNumber } from 'ethers'
+import { App, connect } from '@aragon/connect'
 import { VotingConnectorTheGraph, Vote, Cast } from '../../src'
+import { Action, VoteStatus } from '../types'
 
 const VOTING_SUBGRAPH_URL =
-  'https://api.thegraph.com/subgraphs/name/aragon/aragon-voting-rinkeby-staging'
+  'https://api.thegraph.com/subgraphs/name/aragon/aragon-voting-rinkeby'
```
Rekard0, line 7: the tests did not ran successfully, even after swapping the VOTING_SUBGRAPH_URL as stated in the PR description.

- can you provide more context or instruction how you run test locally?
- Is there a way to run the test successfully without swapping the URL?

